PR #25, opened 2019-01-15
Uploading assignment 1 - Inventory Management

kmonster462:
Not quite finished, ran out of time. Used patch from unittest.mock instead of MagicMock to test input.

### DIFF
--- a/students/Kate_Koehn/Lesson01/Assignment/test_file.py
+++ b/students/Kate_Koehn/Lesson01/Assignment/test_file.py
@@ -1,0 +1,104 @@
+from unittest.mock import MagicMock
+from unittest.mock import patch
+from inventory_management.inventory_class import Inventory
+from inventory_management.furniture_class import Furniture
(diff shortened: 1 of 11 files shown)


students/Kate_Koehn/Lesson01/Assignment/test_file.py
```diff
@@ -0,0 +1,104 @@
...
+        assert func == menu.item_info
+
+
+def test_that_main_menu_returns_exit_program_when_user_input_is_q():
+    """tests that in main menu function, user input of q will return exit_program"""
+    with patch("builtins.input", return_value = "q"):
+        func = menu.main_menu()
+
+        assert func == menu.exit_program
+
```
codedragon: Notice how similar these last three tests are. We didn't really talk about parameterize, but this would be an excellent place to use it. https://docs.pytest.org/en/latest/parametrize.html

PR #27, opened 2019-01-15
lesson01 and 02 assignment submission

habtamua:
`python -m pylint ./inventory_management --include-naming-hint=y`
Your code has been rated at 10.00/10 (previous run: **10.00/10, +0.00**)

`coverage run --source=inventory_management -m unittest; coverage report`

- **_Name                                                                             Stmts   Miss  Cover_**
---------------------------------------------------------------------
- inventory_management/__init__.py                               0      0   100%
- inventory_management/electricappliancesclass.py   11      0   100%
- inventory_management/furnitureclass.py                   11      0   100%
- inventory_management/inventoryclass.py                  13      0   100%
- inventory_management/management.py                    36     27    25%
- inventory_management/market_prices.py                   2      1    50%
- ---------------------------------------------------------------------
- TOTAL                                                                           73     28    62%


students/HABTAMU/lesson01/assignment/tests/test_unit.py
```diff
@@ -0,0 +1,105 @@
...
+class Testmanagementclass(TestCase):
+    def test_get_price(self):
+        item_price = mgm_mod.get_price()
+        pass
+
+
+    def test_item_info(self):
+        pass
+
+    def test_exit_program(self):
```
codedragon: Testing sys exit is unnecessary, so I wouldn't even do a placeholder for it.

students/HABTAMU/lesson01/assignment/tests/test_unit.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+class Testmanagementclass(TestCase):
+    def test_get_price(self):
+        item_price = mgm_mod.get_price()
+        pass
+
+
+    def test_item_info(self):
+        pass
```
codedragon: As written, item_info is hard to test. Imagine testing this instead:
```
def item_info(item):
    if itemCode in fullInventory:
        printDict = fullInventory[itemCode]
        for k,v in printDict.items():
            return (k, v)
    return None
```
How could you fit this into the application?

Note: this code assumes the user only cares about one of any particular item in inventory, which may not be true.

students/HABTAMU/lesson02/assignment/code/charges_calc.py
```diff
@@ -0,0 +1,179 @@
...
+    # parser.add_argument('-d', required=True)
+
+    return parser.parse_args()
+
+
+def load_rentals_file(filename):
+    with open(filename) as file:
+        try:
+            data = json.load(file)
+        except:
```
codedragon: you should save the exception:
```
except Exception as err:
    log.error(err)
```
Even better, figure out what errors you can deal with, and only except those errors.

students/HABTAMU/lesson02/assignment/code/charges_calc.py
```diff
@@ -0,0 +1,179 @@
...
+            # Error loading json from file, for the reason expected property name error out.
+            logging.error("Error loading json from file {}".format(filename))
+            logging.warning(
+                "JSONDecodeError: Expecting property name enclosed in double quotes: line 5884 column 23")
```
codedragon: This may have been the error once, but that does not mean it will be the error for all data.

students/HABTAMU/lesson02/assignment/code/charges_calc.py
```diff
@@ -0,0 +1,179 @@
...
+
+            # Here capture logging stacktrace or traceback to track the code’s execution error.
+            logging.exception("The next few lines are Traceback error message")
+
+            # Error out with exit(1) to failed loading json
+            exit(1)
```
codedragon: Make sure you tell your user why the application is exiting early. Logs are for developers to diagnose problems, users generally don't see them.

students/HABTAMU/lesson02/assignment/code/charges_calc.py
```diff
@@ -0,0 +1,179 @@
...
+            # wrote another script (clean_up.py) to switch rental_start and rental_end.
+            logging.error("rental_start needs to be prior to rental_end, total_days shouldn't be negative")
+
+            value['total_price'] = value['total_days'] * value['price_per_day']
+            value['sqrt_total_price'] = math.sqrt(value['total_price'])
+            if value['units_rented'] == 0:
+                value['unit_cost'] = value['total_price']
+            else:
+                value['unit_cost'] = value['total_price'] / \
+                    value['units_rented']
```
codedragon: This is way too much in a try block. You want to limit what is in a try block to the minimum, so you have some idea where the error came from.

students/HABTAMU/lesson02/assignment/code/clean_up.py
```diff
@@ -0,0 +1,24 @@
...
+     for value in data.values():
+             rental_start = datetime.datetime.strptime(value['rental_start'], '%m/%d/%y')
+             rental_end = datetime.datetime.strptime(value['rental_end'], '%m/%d/%y')
+             if (rental_end - rental_start).days < 0:
+                     temp_value = value['rental_start']
+                     value['rental_start'] = value['rental_end']
+                     value['rental_end'] = temp_value
+
+with open('correct_source.json', 'w', encoding='utf-8') as output:
+     output.write(json.dumps(data, indent=4,ensure_ascii=False))
```
codedragon: I like the approach of having a separate cleaning data module. You could have separate cleaning modules for different customers or data, and then you could assume that your data is clean in your other module, and you can verify that you are doing the right thing with your customer with everything you are doing in one place. Nice! I would make it into a function, however. Makes it easier to test, and you do want tests on it.

PR #28, opened 2019-01-15
Assignment 1

CodeChameleon:
- Some refactoring complete, but I know there is more that I can do depending on how the app needs to expand

- Each file has at least 90% code coverage

- pylint is 10, but I did disable two warnings for too many parameters in two files since I was certain on how to refactor appropriately (maybe kwargs?) and unused parameter in market_prices since we weren't supposed to alter the structure of that file.


students/DrewSmith/lessons/lesson01/assignment/tests/test_unit.py
```diff
@@ -0,0 +1,289 @@
...
+        self.assertEqual(result, "test_key:test_val")
+    
+    @mock.patch('main.management')
+    def test_item_info_no_input(self, mock_management):
+        mock_management.get_item = MagicMock(return_value={"test_key": "test_val"})
+        with mock.patch('main.safe_input', return_value='my_test_value'):
+            with captured_output() as (out, err):
+                main.item_info()
+            result = out.getvalue().strip()
+        self.assertEqual(result, "test_key:test_val")
```
codedragon: Great job with these tests. I think you understand mock pretty well now. :) User interfaces are a pain in the butt to test.

students/DrewSmith/lessons/lesson02/assignment/code/charges_calc.py
```diff
@@ -0,0 +1,159 @@
...
+    Loads standard logging setup
+
+    :param log_level: logging level from the user
+    (1: Error, 2: Warning, 3: Debug)
+    '''
+    log = "%(asctime)s %(filename)s:%(lineno)-3d %(levelname)s %(message)s"
+    debug_dict = {'1': logging.ERROR, '2': logging.WARNING, '3': logging.DEBUG}
+    if logger is None:
+        logger = logging.getLogger()
+    logger.setLevel(debug_dict.get(log_level, logging.CRITICAL))
```
codedragon: This is one way to handle it, but it is setting up logging. I get it is to the console and shouldn't print anything, since there aren't any critical messages, but why not just return early if we aren't logging anything?

students/DrewSmith/lessons/lesson02/assignment/code/charges_calc.py
```diff
@@ -0,0 +1,159 @@
...
+
+    :param filename: file name to load
+    '''
+
+    result = None
+    try:
+        with open(filename) as file:
+            try:
+                result = json.load(file)
+                logging.debug(f"File loaded successfully: {filename}")
```
codedragon: It is good to limit what is in a try block, and there is such a thing as an else block in a try statement, for this very purpose.
```
try:
   result = json.load(file)
except json.decoder.JSONDecodeError as json_decode_error:
   print(f"Error occurred while reading the JSON file")
   logging.error(f"Error loading JSON: {filename}: " + f"{str(json_decode_error)}")
else:
    logging.debug(f"File loaded successfully: {filename}")
```
I like to think of it as "else if nothing goes wrong"

students/DrewSmith/lessons/lesson02/assignment/code/charges_calc.py
```diff
@@ -0,0 +1,159 @@
...
+            except json.decoder.JSONDecodeError as json_decode_error:
+                print(f"Error occurred while reading the JSON file")
+                logging.error(f"Error loading JSON: {filename}: " + \
+                    f"{str(json_decode_error)}")
```
codedragon: This is awesome! Very happy to see the very specific error catches. :)

students/DrewSmith/lessons/lesson02/assignment/code/charges_calc.py
```diff
@@ -0,0 +1,159 @@
...
+        except KeyError as key_err:
+            logging.warn(key_log.format(key, key_err.args[0], str(key_err)))
+            print(f"WARNING: {key} missing {key_err.args[0]} value")
+        except ValueError as value_error:
+            logging.warn(f"Invalid data: {key}: {str(value_error)}")
+            print(f"WARNING: {key} has invalid units_rented value")
+        except ZeroDivisionError as zero_error:
+            logging.warn(
+                f"Invalid data: {key}: units_rented is 0: {str(zero_error)}")
+            print(f"WARNING: {key} units_rented cannot be 0")
```
codedragon: Very nice. It is a lot of try/except blocks, but so it goes with sanitizing data.

PR #29, opened 2019-01-15
added lesson01 inventory_assignment folder 

Zach-Cooper:
Had some trouble dealing with Mock and test_ingreation. Still working on a couple things but wanted to submit what I have thus far to show what Im working on. Was able to rework my main.py code to make unit_testing easier when I figure out how to truly mock and input. Will make changes once we go over in class and get my questions answered. Overall fun buy very challenging.
Used this to run pylint and disable snake_case/came_case errors:
python -m pylint --disable=C0103,C0111 ./inventory_management


students/Zach Cooper/lesson01/inventory_management/main.py
```diff
@@ -0,0 +1,126 @@
...
+import sys
+from inventory_management.market_prices import get_latest_price
+from inventory_management.inventoryClass import Inventory
+from inventory_management.furnitureClass import Furniture
+from inventory_management.electricAppliancesClass import ElectricAppliances
+# import sys
+# import market_prices
+# import inventoryClass
+# import furnitureClass
+# import electricAppliancesClass
```
codedragon: Main is in the same directory as market_prices, inventoryClass, etc. so the commented out imports are really the ones you want. If main was in the directory above then you would need to import them from their directory. Also, you are using them down below as market_prices.get_latest_price, etc, so those imports would not work even if this was in the directory above.

students/Zach Cooper/lesson02/assignment/code/charges_calc.py
```diff
@@ -0,0 +1,149 @@
...
+    logger.addHandler(file_handler)
+
+    if log_level == 1:
+        logger.setLevel(logging.ERROR)
+    elif log_level == 2:
+        logger.setLevel(logging.WARNING)
+    elif log_level == 3:
+        logger.setLevel(logging.DEBUG)
+
+    return logger
```
codedragon: Note you do not have to return logger. You are not using it. Where you are returning None above just write return (it returns None by default) and at the bottom you do not need a return statement at all.

students/Zach Cooper/lesson02/assignment/code/charges_calc.py
```diff
@@ -0,0 +1,149 @@
...
+
+    # # Could also put levels in a dict
+    # log_level = {'0': logging.CRITICAL,
+    #               '1': logging.ERROR,
+    #               '2': logging.WARNING,
+    #               '3': logging.DEBUG}
```
codedragon: Note that if you put it in a dictionary, you do not need all of the conditionals, just
`logger.setLevel(log_dict[log_level])`
You cannot call the dictionary and the input variable both log_level

students/Zach Cooper/lesson02/assignment/code/charges_calc.py
```diff
@@ -0,0 +1,149 @@
...
+    :param filename: source file
+    :return: data
+    """
+    logging.debug("Opening file %s", filename)
+    with open(filename) as file:
+        try:
+            data = json.load(file)
+        except Exception as err:
+            logging.error("Error reading input file %s", file)
+            print(f"Error reading the input file: {err}")
```
codedragon: Good. Even better is play around and find out what errors you are likely to get and only except those.

students/Zach Cooper/lesson02/assignment/code/charges_calc.py
```diff
@@ -0,0 +1,149 @@
...
+            print(f"Date format is not correct. Should be m/d/y.")
+            logging.debug(value)
+
+        # Total Days Field
+        value['total_days'] = (rental_end - rental_start).days
+        if value["total_days"] < 0:
+            logging.error("Total days is negative: %s", value)
+            print("Total days should not be negative. Rental start and rental end"
+                  "dates need to be checked.")
+            logging.debug(value)
```
codedragon: Printing is good, but since you aren't stopping the loop, they won't know which value was the problem. I would add to the print statement the value, so they don't have to search through all of the data for the negative numbers.

students/Zach Cooper/lesson02/assignment/code/charges_calc.py
```diff
@@ -0,0 +1,149 @@
...
+    :param filename: output filename
+    :param data: formatted data
+    :return:
+    """
+    logging.debug("save_to_json")
+    logging.debug("Writes updated json file as outfile")
+    with open(filename, 'w') as file:
+        if data is None:
+            logging.warning("No data is outputed")
+        json.dump(data, file, indent=4)
```
codedragon: Good catch, but might as well check to see if data is None before you open the file. :)

PR #33, opened 2019-01-16
Assignment01: Going for the partial credit option

dancornutt:
Had trouble completing the Mock portial and getting the python packages to work with main.py outside inventory_management


students/DanCornutt/lessons/lesson02/assignment/code/charges_calc.py
```diff
@@ -0,0 +1,116 @@
...
+    for value in data.values():
+        logs.debug("Calculating fields for {} item".format(
+            value['product_code']))
+        try:
+            rental_start = datetime.datetime.strptime(value['rental_start'],
+                                                      '%m/%d/%y')
+            if value['rental_end'] == "":
+                logs.warning('rental_end is null')
+                logs.error('setting rental_end date to now')
+                rental_end = datetime.datetime.now()
```
codedragon: Lol, you might have just cost someone a lot of money. ;) This is a tough one, and really we shouldn't be guessing, but as long as we log what we do and LET THE USER know, it is okay.

PR #34, opened 2019-01-16
lesson01 assignment

rushinstuffin:
Lesson one assignment. Needs more work. Tried to fix the fact that I pushed changes that were for the main directory, and it looks like I did.


students/KyleBarry/lesson01/assignment/inventory_management/test_unit.py
```diff
@@ -0,0 +1,61 @@
...
+
+class MarketPriceTests(TestCase):
+
+    def test_market_price(self):
+        result = get_latest_price("")
+
+        self.assertEqual(24, result)
+
+class MainTests(TestCase):
+
```
codedragon: Here is an example of mock generating input values. This is assuming that there are 3 inputs in this function, which isn't true, but it is an example:
```
with mock.patch('main.input', side_effect=['13','chair','3']):
            add_new_item()
```
If you only need to fill in one input, you can do return_value = '4353', for example, but if there are multiple inputs, you need to use side_effect with a list. Each time your test hits an input in the code that is running, it will submit the next item in the list.

students/KyleBarry/lesson03/test_customer.py
```diff
@@ -0,0 +1,70 @@
...
+def test_update_customer_credit():
+
+    update_customer_credit("000001", 10000)
+    try:
+        cust = Customer.get(Customer.customer_id == "000001")
+        assert cust.credit_limit == 10000
+    except Exception as e:
+        logging.info(e)
+        raise ValueError
+        assert True
```
codedragon: You can just type raise instead of raise ValueError. If you just do raise, it will re-raise the original exception. No reason to assert True in your except block. For that matter, you shouldn't put the try block in your test at all. If the test errors, you know you have to fix something.

PR #35, opened 2019-01-16
Assignment 1 & 2

Misseswaddles:


Misseswaddles:
Adding Assignment 2

students/cfarris/lessons/lesson02/assignment/code/charges_calc.py
```diff
@@ -0,0 +1,125 @@
...
+            rental_start = datetime.datetime.strptime(value['rental_start'], '%m/%d/%y')
+        except ValueError as err:
+            logging.error(err)
+            print('Error with rental_start', value['rental_start'])
+
+        try:
+            rental_end = datetime.datetime.strptime(value['rental_end'], '%m/%d/%y')
+        except ValueError as err:
+            logging.error(err)
+            print('error with rental_end', value['rental_end'])
```
codedragon: except ValueError as err:
    logging.error(err)
   raise

students/cfarris/lessons/lesson03/assignment/customer_model.py
```diff
@@ -0,0 +1,71 @@
...
+"""
+    Simple database example with Peewee ORM, sqlite and Python
+    Here we define the schema
+    Use logging for messages so they can be turned off
+
+"""
+import logging
+from peewee import *
+
+logging.basicConfig(level=logging.INFO)
```
codedragon: You don't have to set the basicConfig more than once, and probably better not to.

students/cfarris/lessons/lesson03/assignment/test_basic_operations.py
```diff
@@ -0,0 +1,76 @@
+#!/usr/bin/env Python3
+import pytest
+from basic_operations import add_customer, search_customer, delete_customer
+from customer_model import Customer
+
+##note, if you get an integrity error be sure clear your database.
```
codedragon: Would, in fact, be good to have some test fixtures that create a database at the beginning and tear it down at the end. :)

students/cfarris/lessons/lesson03/assignment/basic_operations.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+if __name__ == '__main__':
+
+    #added customer below already
+    #add_customer('2212','Jerry','Blank','19202 130th Ave, Sun City, Az. 98125', '123-456-6543', 'jb@aol.com', False, 500.00)    
+    print("already added customers. name == main is just a placeholder")
+    search_customer('2212')
+    search_customer('1')
+    #delete_customer('2212') #works correctly
```
codedragon: The if name == main block might be a good place to check to see if the database exists, and create as necessary. Probably not how the customer will interact with our api, and won't get run by our tests, but can use for simple interactions and playing around.

PR #36, opened 2019-01-16
added files for lesson01 assignment

VioletLuo:
the assignment is not finished yet, still has importing issues

### DIFF
--- a/students/ZidanLuo/lessons/lesson01/Assignments/Assignment01/test_unit.py
+++ b/students/ZidanLuo/lessons/lesson01/Assignments/Assignment01/test_unit.py
@@ -1,0 +1,95 @@
+"""
+This is a unit test for all classes in the inventory management system
+"""
+from unittest import TestCase
+from unittest.mock import MagicMock
+
+
+from inventory_management.inventory_class import Inventory
+from inventory_management.furniture_class import Furniture
+from inventory_management import market_prices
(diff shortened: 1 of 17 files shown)


students/ZidanLuo/lessons/lesson01/Assignments/Assignment01/test_unit.py
```diff
@@ -0,0 +1,107 @@
...
+        self.assertEqual(result['voltage'], '110v')
+
+
+class MainMenuTests(TestCase):
+    """
+    This class tests the methods in the Main module
+    """
+    @patch('main.main_menu',return_value=add_new_item)
+    def test_get_price(self, main_menu):
+        self.assertEqual(main_menu(1), add_new_item)
```
codedragon: I think the problem is a misunderstanding about how patch works. The thing that you want to patch is the input function, so that is what goes in the patch call, and the return_value that you are faking is "1", since that is what the user would be entering into the input command if we were running this on the command line. However, because of your clever use of user_prompt, you don't actually even need to fake anything. Here are two different ways to test this function for input 1, that should both work:
```
    @patch('main.input',return_value="1")
    def test_get_price(self, fake_input):
        self.assertEqual(main_menu(), add_new_item)

    def test_get_price_directly(self):
        self.assertEqual(main_menu("1"), add_new_item)
```
Hope this helps!

PR #37, opened 2019-01-19
Submitting week 2 assignment - Charges_calc.py

kmonster462:
Submitting files for week 2 assignment.


students/Kate_Koehn/Lesson02/charges_calc.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+def set_log_level(log_level=0):
+    """
+    function to set level of logging desired from command line when running program
+    :param log_level: 0, 1, 2, or 3
+    :return: returns different level of logging when running the program
+    """
+    if log_level == 0:
+        return
```
codedragon: Notice that all of the logging stuff at the top of the file could be below this return as well. Even less overhead when not logging!

Issue #2, opened 2019-01-01
Write access needed to master for me


pythonsavvy:
Please can someone grant - thanks

PythonCHB:
Odd -- not sure how you were able to create the repo, but not be able to push to it.

But I've made you and "owner" of this org, so you should be god to go.

Is this intended to be a template for the 220 classes?, or one for a particular class? If the latter, then se may want to change the name to include the modality and timeframe: e.g. Winter2019Classroom or something.

Also, you could create a team for this class -- that would be the instructor and TA.

pythonsavvy:
Thanks - I think its cos I got the org wrong originally. This is the student's template for py220 activities and assignments. About to push the content; its been in gitlab. Tis has all the autograding with pytest, linitng and coverage

PR #39, opened 2019-01-22
Assignment #2

vanulor:
Log,code,output file


students/BrianNisonger/Lesson2/assignment/charges_calc.py
```diff
@@ -0,0 +1,125 @@
...
+def configure_logging(debug_level):
+    """
+    Configures the logging
+    """
+    format_logger = '%(asctime)s %(filename)s:%(lineno)-3d%(levelname)s'
+    format_logger += ' %(message)s'
+    log_file = datetime.datetime.now().strftime('%Y-%m-%d') + '.log'
+    debug_level = int(debug_level)
+    if debug_level == 0:
+        logging.basicConfig(level=logging.CRITICAL)
```
codedragon: If you're not going to log, no point in all of the logging overhead. Note that if someone did stick in a critical log message it would show up in the console.

students/BrianNisonger/Lesson2/assignment/charges_calc.py
```diff
@@ -0,0 +1,125 @@
...
+    Loads json file of rental data
+    """
+    logging.debug("load_rentals_file")
+    logging.debug("Opens a json file and reads it in")
+    with open(filename) as file:
+        try:
+            data1 = json.load(file)
+        except Exception as error_msg:
+            logging.error(error_msg)
+            data1 = None
```
codedragon: If data is None, there is no point in continuing. Tell the user there was a problem with the file, and exit.

students/BrianNisonger/Lesson2/assignment/charges_calc.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+def calculate_additional_fields(data_new):
+    """
+    Calculates additional fields from raw property data
+    """
+    logging.debug("calculate_additional_fields")
+    if data_new:
+        for value in data_new.values():
+            try:
```
codedragon: Try to limit what is in a try block, so you can easily figure out where the error came from. In this case, it means a bunch of try/except blocks, but that is normal when sanitizing data.

students/BrianNisonger/Lesson2/assignment/charges_calc.py
```diff
@@ -0,0 +1,125 @@
...
+                warning_string = f"Incorrect date for {value['product_code']}"
```
codedragon: This is a great idea. We often put strings like this in a config so they can be re-used lots of places, and checked in tests.

students/BrianNisonger/Lesson2/assignment/charges_calc.py
```diff
@@ -0,0 +1,125 @@
...
+                if value["rental_start"] == '' or value["rental_end"] == '':
+                    rental_start = 0
+                    rental_end = rental_start
+                    logging.warning(warning_string)
+                    logging.debug(value)
```
codedragon: This is going to put the rental_start in the wrong initial format, and neither get formatted using datetime.datetime.strptime. It's a tough problem, and no matter what you do you have to talk to your customer. As a stop-gap, I would probably give them both the same date, and then not put the datetime.datetime.strptime in the else block.

PR #41, opened 2019-01-27
Assignment 3

ktbzzz:


students/JaeKim/lesson03/assignment/basic_operations.py
```diff
@@ -0,0 +1,108 @@
...
+    :return:
+    """
+    try:
+        if stage == 'prod':
+            results = Customer.get(cm.customer_id == customer_id)
+
+        if stage == 'dev':
+            results = TestCustomer.get(TestCustomer.customer_id == customer_id)
+    except Exception:
+        return False
```
codedragon: You should log your exceptions, especially because you are catching all of them. An exception that you aren't expecting could happen, and you would never know.

students/JaeKim/lesson03/assignment/basic_operations.py
```diff
@@ -0,0 +1,108 @@
...
+    :param name:
+    :param lastname:
+    :param home_address:
+    :param phone_number:
+    :param email_address:
+    :param status:
+    :param credit_limit:
+    :param stage:
+    :return:
+    """
```
codedragon: Cool, nice doc string. Could also include any limitations on the data (status True/False or no $ in money. Whatever type you want for each variable.

students/JaeKim/lesson03/assignment/test_basic_operations.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+def teardown_test_db():
+    """
+    Drops test_customer, used for teardown
+    """
+    cm.database.drop_tables([
+        cm.TestCustomer
+    ])
+    cm.database.close()
```
codedragon: Great job with the setup and teardown methods. If you use pytest fixtures, you don't have to put them in every test, it automatically runs them for you. 
https://docs.pytest.org/en/latest/fixture.html

students/JaeKim/lesson03/assignment/test_basic_operations.py
```diff
@@ -0,0 +1,185 @@
...
+    assert valid_customer.first_name == OK_CUSTOMER['name']
+    assert valid_customer.customer_id == OK_CUSTOMER['customer_id']
+    assert valid_customer.last_name == OK_CUSTOMER['lastname']
+    assert valid_customer.home_address == OK_CUSTOMER['home_address']
+    assert valid_customer.phone_number == OK_CUSTOMER['phone_number']
+    assert valid_customer.email_address == OK_CUSTOMER['email_address']
+    assert valid_customer.status == OK_CUSTOMER['status']
+    assert valid_customer.credit_limit == OK_CUSTOMER['credit_limit']
+
+    teardown_test_db()
```
codedragon: You don't have to do the teardown method here, since you do it in your setup method.

students/JaeKim/lesson03/assignment/test_basic_operations.py
```diff
@@ -0,0 +1,185 @@
...
+    assert valid_customer.email_address == OK_CUSTOMER['email_address']
+    assert valid_customer.status == OK_CUSTOMER['status']
+    assert valid_customer.credit_limit == OK_CUSTOMER['credit_limit']
+
+    delete_customer(OK_CUSTOMER['customer_id'], stage='dev')
+    valid_customer = search_customer(OK_CUSTOMER['customer_id'], stage='dev')
+
+    assert valid_customer is False
+
+    teardown_test_db()
```
codedragon: This always feel a bit weird, doing in one test what another has already tested. You don't have to make sure it is there after the add in this test. You tested that elsewhere; here you are testing delete, so just make sure it is gone after the delete. But, sometimes I do skip the add tests, since it gets tested so much in all of the other tests.

PR #42, opened 2019-01-29
Added activity and assignment files under lesson03

dtingg:
My tests for the basic_operations file work, but they take a really long time.  Is there a better way to create a testing framework using a sample database?


students/DiannaTingg/lessons/lesson03/assignment/create_customers.py
```diff
@@ -0,0 +1,19 @@
...
+    logging.basicConfig(level=logging.INFO)
+    logger = logging.getLogger(__name__)
+
+    logger.info('Creating a database using the customer model.')
+
+    cm.database.create_tables([
+        cm.Customer
+    ])
+
+    cm.database.close()
```
codedragon: I think you want to wait to close the connection until you are done. This might be part of the problem with your tests, you are closing the connection after every transaction, so peewee has to re-establish a connection every time.

students/DiannaTingg/lessons/lesson03/assignment/test_basic_operations.py
```diff
@@ -0,0 +1,190 @@
...
+    bo.add_customer(**customer2)
+
+    test_customer1 = cm.Customer.get(cm.Customer.customer_id == "1")
+    assert test_customer1.first_name == customer1["first_name"]
+    assert test_customer1.last_name == customer1["last_name"]
+
+    test_customer2 = cm.Customer.get(cm.Customer.customer_id == "2")
+    assert test_customer2.first_name == customer2["first_name"]
+    assert test_customer2.last_name == customer2["last_name"]
+
```
codedragon: If one customer works, the other should, so just do one. This is something you might cover in an integration test, no need to here.

students/DiannaTingg/lessons/lesson03/assignment/test_basic_operations.py
```diff
@@ -0,0 +1,190 @@
...
+
+def test_add_customers_duplicate():
+    create_empty_database()
+    bo.add_customer(**customer1)
+
+    with pytest.raises(pw.IntegrityError):
+        bo.add_customer(**customer1)
+
+    clear_database()
```
codedragon: Great. Easy to forget the tests looking for errors. You might consider raising a more friendly for your user error instead (although I ran into problems with testing doing this with peewee).

students/DiannaTingg/lessons/lesson03/assignment/test_basic_operations.py
```diff
@@ -0,0 +1,190 @@
...
+
+    cm.database.create_tables([
+        cm.Customer
+    ])
+    bo.add_customer(**customer1)
+    bo.add_customer(**customer2)
+    bo.add_customer(**customer3)
+    bo.add_customer(**customer4)
+
+    cm.database.close()
```
codedragon: You could just use this function for all of your tests; no need to have some tests start with an empty database.

students/DiannaTingg/lessons/lesson03/assignment/test_basic_operations.py
```diff
@@ -0,0 +1,190 @@
...
+
+    assert test_customer["email_address"] == customer3["email_address"]
+
+    clear_database()
+
+
+def test_search_customer_missing():
+    create_sample_database()
+
+    assert bo.search_customer("8") == dict()
```
codedragon: slightly less typing to do == {}

students/DiannaTingg/lessons/lesson03/assignment/test_basic_operations.py
```diff
@@ -0,0 +1,190 @@
...
+
+    updated_customer = cm.Customer.get(cm.Customer.customer_id == 2)
+    assert updated_customer.credit_limit == 300.00
+
+    # Search for a customer
+    target_customer = bo.search_customer(customer4["customer_id"])
+    assert target_customer["email_address"] == customer4["email_address"]
+
+    # List active customers
+    assert bo.list_active_customers() == 1
```
codedragon: Good test, this is the idea for a lifecycle test. I may not have been clear about the differences. I'll fix that next Tuesday. At any rate, I'm not sure I'd bother with integration tests yet, since there isn't much being integrated. When something is actually using the functions in basic_operations, that is probably time for integration tests.

students/DiannaTingg/lessons/lesson03/assignment/test_basic_operations.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+def create_empty_database():
+    clear_database()
+
+    cm.database.create_tables([
+        cm.Customer
+    ])
+
+    cm.database.close()
```
codedragon: Consider making these two functions fixtures. Less typing, it automatically runs them before/after each test.
https://docs.pytest.org/en/latest/fixture.html

PR #43, opened 2019-01-29
submit lesson3 hw

ShujinCao:


students/ShujinCao/lesson03/assignment/basic_operations.py
```diff
@@ -0,0 +1,59 @@
...
+from customer_model import Customer
+from peewee import fn
+
+def add_customer(customer_id, name, last_name, home_address, phone_number,
+                 email_address, status, credit_limit):
+    try:
+        credit_limit = float(credit_limit)
+    except ValueError as err:
+        logging.error(err)
+        raise
```
codedragon: Yes, it seems peewee isn't checking types. Looks good. Note that since you know what is likely to cause this error ($), you might customize the error message to say they shouldn't use $ or other things that are not numbers.

students/ShujinCao/lesson03/assignment/basic_operations.py
```diff
@@ -0,0 +1,59 @@
...
+        newn_customer = Customer.get(Customer.customer_id == customer_id)
+        cstm_dict = {'name': newn_customer.name,
+                     'last_name': newn_customer.last_name,
+                     'email_address': newn_customer.email_address,
+                     'phone_number': newn_customer.phone_number}
+    except Exception as e:
+        cstm_dict = {}
+        return cstm_dict
+
+    return cstm_dict
```
codedragon: You don't actually need two return statements here. One outside of the try/except block will return either the empty dict or the full one, depending on where it was created.

students/ShujinCao/lesson03/assignment/test_basic_operations.py
```diff
@@ -0,0 +1,86 @@
...
+                'home_address': 'Beverly Hills Plaza, Compton, Ca. 90210',
+                'phone_number': '198-765-4321',
+                'email_address': 'Jenny_Igotyonumber@gmail.com',
+                'status': True,
+                'credit_limit': 40.00}
+
+def test_add_customer():
+    # add_customer(**good_customer)
+    test_customer = Customer.get(Customer.customer_id==good_customer['customer_id'])
+    assert test_customer.email_address == good_customer['email_address']
```
codedragon: I'm thinking that you commented out the add_customer, because the customer was still there from the previous run. You should make fixtures for before/after your tests that create your table before the run and delete it after. https://docs.pytest.org/en/latest/fixture.html You can look in my student directory for an example of doing it with unittest.

students/ShujinCao/lesson03/assignment/test_basic_operations.py
```diff
@@ -0,0 +1,86 @@
...
+
+def test_delete_customer():
+    '''
+    test delete func
+    '''
+    add_customer(**good_customer2)
+    test_customer = Customer.get(Customer.customer_id == good_customer2['customer_id'])
+    assert test_customer.email_address == good_customer2['email_address']
+    deleted = delete_customer(test_customer.customer_id)
+    assert deleted == 'customer deleted'
```
codedragon: Probably better to check the database to see if it is there. One person solved this by counting how many people were in the database before and after you delete.

students/ShujinCao/lesson03/assignment/basic_operations.py
```diff
@@ -0,0 +1,59 @@
...
+    except Exception as e:
+        print('Customer wasn\'t found.')
+    return 'customer deleted'
+
+def update_customer_credit(customer_id, credit_limit):
+    try:
+        u = Customer.update(credit_limit=credit_limit).where(Customer.customer_id == customer_id)
+        u.execute()
+    except ValueError:
+        print('Customer wasn\'t found.')
```
codedragon: I'm not sure it is a ValueError if the customer isn't found here. Might want to check/test that.

PR #45, opened 2019-01-29
Adding my completed code for Lesson 3. 

vanulor:
Adding code and tests.  If I have additional time I will add more tests.  Everything runs and all tests pass.

### DIFF
--- a/students/BrianNisonger/Lesson3/assignment/basic_operations.py
+++ b/students/BrianNisonger/Lesson3/assignment/basic_operations.py
@@ -1,0 +1,66 @@
+from peewee import *
+from customer_model import Customer
+from customer_model import database
+
+
+def model_to_dict(model):
+    model_dict = {
+        'customer_id': model.customer_id,
+        'name': model.first_name,
+        'last_name': model.last_name,
+        'home_address': model.home_address,
+        'phone_number': model.phone_number,
+        'email_address': model.email_address,
+        'status': model.status,
+        'credit_limit': model.credit_limit
+    }
+    return model_dict
+
+
+def add_customer(customer_id, name, last_name, home_address, phone_number,
+                 email_address, status, credit_limit):
+    try:
+        credit_limit = float(credit_limit)
+    except ValueError as err:
+        raise
+    if not database.table_exists('Customer'):
+        database.create_tables([Customer])
+    new_customer = Customer.create(
+        customer_id=customer_id,
+        first_name=name,
+        last_name=last_name,
+        home_address=home_address,
+        phone_number=phone_number,
+        email_address=email_address,
+        status=status,
+        credit_limit=credit_limit)
+    new_customer.save()
+
+
+def search_customer(customer_id=None):
+    find_customer_model = Customer.get_or_none(
+        Customer.customer_id == customer_id)
+    if find_customer_model:
+        find_customer_dict = model_to_dict(find_customer_model)
+    else:
+        find_customer_dict = {}
+    return find_customer_dict
+
+
+def delete_customer(customer_id=None):
+    delete_customer = Customer.get_or_none(Customer.customer_id == customer_id)
+    if delete_customer:
+        delete_customer.delete_instance()
+
+
+def update_customer_credit(customer_id=None, credit_limit=None):
+    update_customer = Customer.get_or_none(Customer.customer_id == customer_id)
+    if update_customer:
+        update_customer.credit_limit = credit_limit
+        update_customer.save()
+    else:
+        raise ValueError
+
+
+def list_active_customers():
+    return Customer.select().where(Customer.status == True).count()
--- a/students/BrianNisonger/Lesson3/assignment/test_customer_db.py
+++ b/students/BrianNisonger/Lesson3/assignment/test_customer_db.py
@@ -1,0 +1,169 @@
+from basic_operations import add_customer
+from basic_operations import search_customer
+from basic_operations import delete_customer
+from basic_operations import update_customer_credit
+from basic_operations import list_active_customers
+
+from customer_model import Customer
+import pytest
+
+ok_customer = {
+    'customer_id': 'W3434fd',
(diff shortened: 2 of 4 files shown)


students/BrianNisonger/Lesson3/assignment/basic_operations.py
```diff
@@ -0,0 +1,66 @@
...
+        'customer_id': model.customer_id,
+        'name': model.first_name,
+        'last_name': model.last_name,
+        'home_address': model.home_address,
+        'phone_number': model.phone_number,
+        'email_address': model.email_address,
+        'status': model.status,
+        'credit_limit': model.credit_limit
+    }
+    return model_dict
```
codedragon: Good idea. I did notice in the peewee documentation, you can return select statements as dictionaries, but does seem silly to do a select for the one entry you want.

students/BrianNisonger/Lesson3/assignment/basic_operations.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+def search_customer(customer_id=None):
+    find_customer_model = Customer.get_or_none(
+        Customer.customer_id == customer_id)
+    if find_customer_model:
+        find_customer_dict = model_to_dict(find_customer_model)
+    else:
+        find_customer_dict = {}
+    return find_customer_dict
```
codedragon: Very clean.

students/BrianNisonger/Lesson3/assignment/basic_operations.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+def delete_customer(customer_id=None):
+    delete_customer = Customer.get_or_none(Customer.customer_id == customer_id)
+    if delete_customer:
+        delete_customer.delete_instance()
```
codedragon: Great way around having to check for exceptions.

students/BrianNisonger/Lesson3/assignment/basic_operations.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+def update_customer_credit(customer_id=None, credit_limit=None):
+    update_customer = Customer.get_or_none(Customer.customer_id == customer_id)
+    if update_customer:
+        update_customer.credit_limit = credit_limit
+        update_customer.save()
+    else:
+        raise ValueError
```
codedragon: As long as your raising an error, you should tell them what happened:
raise ValueError("Unable to update credit, did you use only numbers?")

students/BrianNisonger/Lesson3/assignment/basic_operations.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+def add_customer(customer_id, name, last_name, home_address, phone_number,
+                 email_address, status, credit_limit):
+    try:
+        credit_limit = float(credit_limit)
+    except ValueError as err:
+        raise
+    if not database.table_exists('Customer'):
+        database.create_tables([Customer])
```
codedragon: This is a clever idea, but it assumes you always start with an add. Which isn't terrible, if you do something else, you get an error and realize there is no table, but it isn't necessarily clear to the user that they should then just try to add a customer and the table will just magically appear. And, once there is a UI, you really just want to create the table when you start the ui. So, I'd have a special create table function.

students/BrianNisonger/Lesson3/assignment/basic_operations.py
```diff
@@ -0,0 +1,66 @@
...
+    }
+    return model_dict
+
+
+def add_customer(customer_id, name, last_name, home_address, phone_number,
+                 email_address, status, credit_limit):
+    try:
+        credit_limit = float(credit_limit)
+    except ValueError as err:
+        raise
```
codedragon: If you are going to just raise it again, why catch it at all?

students/BrianNisonger/Lesson3/assignment/test_customer_db.py
```diff
@@ -0,0 +1,169 @@
...
+    test_customer3 = Customer.get(
+        Customer.customer_id == third_customer["customer_id"])
+    test_customer3.status = True
+    test_customer3.save()
+    active_customers = list_active_customers()
+    assert active_customers == 3
+    clean_up(ok_customer['customer_id'])
+    clean_up(second_customer['customer_id'])
+    clean_up(third_customer['customer_id'])
+    clean_up(fourth_customer['customer_id'])
```
codedragon: This should probably be two separate tests. This is what I would call a lifecycle test, where you do a bunch of different things as a user might do, and make sure it all works in one go.

students/BrianNisonger/Lesson3/assignment/test_customer_db.py
```diff
@@ -0,0 +1,169 @@
+from basic_operations import add_customer
+from basic_operations import search_customer
+from basic_operations import delete_customer
+from basic_operations import update_customer_credit
+from basic_operations import list_active_customers
+
+from customer_model import Customer
+import pytest
+
```
codedragon: Nice fixtures. Pytest has fixtures you can use to make life easier. https://docs.pytest.org/en/latest/fixture.html

students/BrianNisonger/Lesson3/assignment/test_customer_db.py
```diff
@@ -0,0 +1,169 @@
...
+
+def test_search_customer_null():
+    add_customer(**ok_customer)
+    customer_dict = search_customer("12345")
+    assert customer_dict == {}
+    customer_dict = search_customer(7)
+    assert customer_dict == {}
+    customer_dict = search_customer()
+    assert customer_dict == {}
```
codedragon: This would be a good candidate for parameterizing. Pytest also has a parameterize module. Anytime you have more than one assert in a test, you should ask yourself if there is a better way. If one fails, the rest are not run, so you do not get that additional information.

PR #46, opened 2019-01-29
Assignment 3

CodeChameleon:
Besides the basic_operations, test_basic_operations, and customers_model I built out the create_customer to allow the testing module to dictate that the "test" environment should be used. This creates/uses a separate database (customers_test.db) that can be altered without affecting "prod" (customers.db). I'm not sure if this was the right approach.


students/DrewSmith/lessons/lesson03/assignment/basic_operations.py
```diff
@@ -0,0 +1,86 @@
...
+    ''' Add a customer '''
+    try:
+        credit_limit = float(credit_limit)
+    except ValueError as err:
+        logging.error(err)
+        raise
+
+    if not isinstance(status, bool):
+        logging.error(f"Invalid value: status is not a bool: {status}")
+        raise ValueError(f"Invalid value: status is not a bool: {status}")
```
codedragon: Nice. I like f strings. Note you are writing the same thing twice. You could set it to a variable instead. This is the sort of thing we keep in a config file, that way you change it one place and everyone sees it, including tests.

students/DrewSmith/lessons/lesson03/assignment/basic_operations.py
```diff
@@ -0,0 +1,86 @@
...
+    Updates the customer credit amount
+
+    :param customer_id: customer id to search and update
+    :param credit_limit: credit limit value to change to
+    '''
+    try:
+        credit_limit = float(credit_limit)
+    except ValueError as err:
+        logging.error(err)
+        raise
```
codedragon: You could try making this error more explicit. You know the sorts of things that will come up, so you could do 
```
raise ValueError("credit limit needs to be a number and only a number (no $)")
```

students/DrewSmith/lessons/lesson03/assignment/create_customers.py
```diff
@@ -0,0 +1,44 @@
...
+import logging
+import sqlite3
+from sqlite3 import Error
+import peewee as pw
+
+
+logging.basicConfig(level=logging.INFO)
+LOGGER = logging.getLogger(__name__)
+
+DATABASE = None
```
codedragon: Database shouldn't be a constant (it is a database represented as a python object, which is not constant).  I like that you have come up with a method to have two different databases. Much better than overwriting the database when you run your tests. :) You can look at my example to see how to do it with a little less code.

students/DrewSmith/lessons/lesson03/assignment/customers_model.py
```diff
@@ -0,0 +1,34 @@
...
+    customer_id = pw.CharField(primary_key=True, max_length=30)
+    first_name = pw.CharField(max_length=30)
+    last_name = pw.CharField(max_length=30)
+    home_address = pw.CharField(max_length=100)
+    phone_number = pw.CharField(max_length=20)
+    email_address = pw.CharField(max_length=100)
+    status = pw.BooleanField()
+    credit_limit = pw.FloatField()
+
+cc.DATABASE.create_tables([Customer])
```
codedragon: You probably don't want this in the global space. For example, in your tests this will create the tables once when you load this file and then again when you start your tests. You want this to just happen once when you start your app, so you probably want it in a __name__ == "__main__" block so it doesn't run every time the file is loaded.

students/DrewSmith/lessons/lesson03/assignment/tests/test_basic_operations.py
```diff
@@ -0,0 +1,132 @@
...
+    cc.DATABASE.create_tables([Customer])
+    
+    for customer_id in customer_id_list:
+        copy = dict(customer_data)
+        copy["customer_id"] = customer_id
+        bo.add_customer(**copy)
+
+    yield
+
+    cc.DATABASE.drop_tables([Customer])
```
codedragon: Great fixture. You actually don't need the drop at the end, since you have it at the beginning.
I don't think you need to copy your dictionary. You aren't using it for anything, just send it in, as is.

students/DrewSmith/lessons/lesson03/assignment/tests/test_basic_operations.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+def test_add_customer_ok():
+    ''' Add a new valid customer '''
+    customer_id = "add_ok"
+    ok_customer = dict(customer_data)
+    ok_customer['customer_id'] = customer_id
+    bo.add_customer(**ok_customer)
```
codedragon: Yes, this is where it is important to make a copy.

students/DrewSmith/lessons/lesson03/assignment/tests/test_basic_operations.py
```diff
@@ -0,0 +1,132 @@
...
+    ''' test list_active_customers count '''
+    result = bo.list_active_customers()
+    assert result == len(customer_id_list)
+    customer = Customer.get(Customer.customer_id == "active_customers")
+    customer.status = False
+    customer.save()
+
+    result = bo.list_active_customers()
+    assert result == len(customer_id_list) - 1
+
```
codedragon: Great testing. :)

PR #47, opened 2019-01-29
Assignment #3

Misseswaddles:
I went through the linter and only left a couple of remaining issues: too many arguments passed, or lines too long (when readability made more sense to keep them longer).

### DIFF
--- a/students/cfarris/lessons/lesson03/assignment/basic_operations.py
+++ b/students/cfarris/lessons/lesson03/assignment/basic_operations.py
@@ -1,16 +1,17 @@
 #!/usr/bin/env Python3
 
-####Assignment 3###
-##basic operations for peewee
+'''
+This scripts provides the functions to interact with
+the sqlite3 database via peewee
+'''
 
-#Current progress: 
-#
 
 import logging
 from customer_model import Customer
 
 
-def add_customer(customer_id, first_name, last_name, home_address, phone_number, email_address, status, credit_limit):
+def add_customer(customer_id, first_name, last_name, home_address,
+                 phone_number, email_address, status, credit_limit):
     '''
      Adds customer to customer database
      PARAMS: Multiple fields. customer_id is primary key
@@ -49,9 +50,9 @@ def search_customer(customer_id):
                      'last_name': retr_customer.last_name,
                      'email_address': retr_customer.email_address,
                      'phone_number': retr_customer.phone_number}
-    except Exception as e:
-        logging.error(e)
-        print('Error occurred retrieving cust: ', e)
+    except Exception as err:
+        logging.error(err)
+        print('Error occurred retrieving cust: ', err)
         cust_dict = {}
         return cust_dict
     return cust_dict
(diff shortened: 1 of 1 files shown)


students/cfarris/lessons/lesson03/assignment/basic_operations.py
```diff
@@ -78,14 +79,35 @@ def update_customer_credit(customer_id, credit_limit):
...
+        retr_customer = Customer.get(Customer.customer_id == customer_id)
+        retr_customer.credit_limit = credit_limit
+        retr_customer.save()
+        if credit_limit < 0:
+            logging.warning('credit_limit parameter is ',
+                            'less than 0: ', credit_limit)
+    except ValueError as err:
+        logging.error('error replacing customers credit limit: ,', err)
+        error_statement = 'An error occured while replacing customers credit.\
+                         Credit unchanged.'
```
codedragon: You can't be sure it wasn't changed. What happens when the error occurred when comparing credit_limit to 0? In that case, the error happened after the credit was saved.

students/cfarris/lessons/lesson03/assignment/basic_operations.py
```diff
@@ -78,14 +79,35 @@ def update_customer_credit(customer_id, credit_limit):
...
 
 if __name__ == '__main__':
-
-    #added customer below already
-    #add_customer('2212','Jerry','Blank','19202 130th Ave, Sun City, Az. 98125', '123-456-6543', 'jb@aol.com', False, 500.00)    
-    print("already added customers. name == main is just a placeholder")
-    search_customer('2212')
-    search_customer('1')
-    #delete_customer('2212') #works correctly
+    print('basic_operations.py is ready to be used with customers.db')
```
codedragon: I think this file just doesn't need an if __name__ == "__main__" block.

students/cfarris/lessons/lesson03/assignment/customer_model.py
```diff
@@ -7,7 +7,7 @@
 
 """
 import logging
-from peewee import *
+from peewee import Model, CharField, BooleanField, FloatField, SqliteDatabase
```
codedragon: Thank goodness. That asterisk hurts my eyes. ;-)

students/cfarris/lessons/lesson03/assignment/test_basic_operations.py
```diff
@@ -1,11 +1,16 @@
 #!/usr/bin/env Python3
+
+'''
+This pytest suite tests basic functionality of basic_operations.py
+'''
+
 import pytest
-from basic_operations import add_customer, search_customer, delete_customer
+from basic_operations import add_customer, search_customer, delete_customer, update_customer_credit, list_active_customers
```
codedragon: This one you probably do want to split into two lines. Or you can do like I did and just import basic_operations as bs or bo, and use them bo.add_customer, etc. Which is kind of fun.

students/cfarris/lessons/lesson03/assignment/test_basic_operations.py
```diff
@@ -64,13 +82,38 @@ def test_search_customer_not_found():
...
+    Test that list_active customers returns the amount of those active.
+    First get an accurate count of active customers,
+    add a new customerthat is active
+    Test to see that the active customers number has increased.
+    '''
+    num_active = list_active_customers()
+    add_customer(**OK_CUSTOMER_3)  # bad customer should work
+    num_active_2 = list_active_customers()
+    delete_customer(OK_CUSTOMER_3['customer_id'])
+    assert (num_active + 1) == num_active_2
```
codedragon: This test is a bit confusing. Why are you deleting ok_customer_3?

PR #48, opened 2019-01-29
Lesson 03 Assignment

Zach-Cooper:
Couple things that I can't seem to figure out with regards to runnig my tests. Could you run pytest on my basic_operations.py and take a look. I just couldnt figure out what needed to be changed on the failed tests.
Thanks


students/Zach Cooper/lesson03/basic_operations.py
```diff
@@ -0,0 +1,103 @@
...
+                                  "phone_number": find_customer.phone_number}
+
+        LOGGER.info('%s %s is in the database. There phone # is %s',
+                    find_customer.first_name, find_customer.last_name,
+                    find_customer.phone_number)
+
+    except pw.DoesNotExist:
+        LOGGER.info('No customer with customer id of %s', Customer.customer_id)
+
+        return find_customer_dict
```
codedragon: I don't think you want your return statement in your except block. This means it is only returned if it is empty. You also have too much in your try block. The only line that will give you a pw.DoesNotExist is the Customer.get line, so that should be the only line in your try. Try has an else block for code that you only want to run when the try succeeds, and then you can return find_customer in all cases. Note that if you have an else block, you don't need the conditional if find_customer, since it will only enter the else if the get succeeds.

students/Zach Cooper/lesson03/basic_operations.py
```diff
@@ -0,0 +1,103 @@
...
+        :param customer_id: customer_id to search and delete for
+    """
+    LOGGER.info("Creating new customer named Tom and then deleting him")
+    try:
+        customer_remove = Customer.get(Customer.customer_id == customer_id)
+        customer_remove.delete_instance()
+
+    except pw.DoesNotExist:
+        return None
+    LOGGER.info("Customer successfully deleted in database")
```
codedragon: Also true in the rest of these. The only thing that will give a does not exist is the get.

students/Zach Cooper/lesson03/test_basic_operations.py
```diff
@@ -0,0 +1,138 @@
...
+    ])
+    cm.database.close()
+
+
+def create_empty_database():
+    clear_database()
+    cm.database.create_tables([
+        cm.Customer
+    ])
+    cm.database.close()
```
codedragon: You don't have to close the connection every test. This only needs to be done at the very end.

students/Zach Cooper/lesson03/test_basic_operations.py
```diff
@@ -0,0 +1,138 @@
...
+    bo.add_customer(**ok_customer2)
+
+    test_customer = cm.Customer.get(cm.Customer.customer_id == ok_customer2['customer_id'])
+    assert test_customer.first_name == ok_customer2['first_name']
+    assert test_customer.last_name == ok_customer2['last_name']
+    assert test_customer.home_address == ok_customer2['home_address']
+    assert test_customer.email_address == ok_customer2['email_address']
+    assert test_customer.status == ok_customer2['status']
+
+    clear_database()
```
codedragon: Look at pytest fixtures, they make setup and teardown easier.

students/Zach Cooper/lesson03/test_basic_operations.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+"""Do I really NEED this"""
+def test_customer_non_existent():
+    create_empty_database()
+
+    assert bo.search_customer('5') == {}
+
+    clear_database()
```
codedragon: It depends. If your api is depending on this code returning an empty dictionary when something is missing (as opposed to raising an error, returning None or something else), you might want to put it there to keep some other developer from deciding that raising an error would be a better idea. A lot of testing is not necessarily about making sure you didn't make a mistake, but about making sure that your promises about how the api work are kept over time.

students/Zach Cooper/lesson03/test_basic_operations.py
```diff
@@ -0,0 +1,138 @@
...
+    with pytest.raises(ValueError):
+        bo.add_customer(**bad_customer)
+
+
+def test_search_customer():
+    create_empty_database()
+    bo.add_customer(**ok_customer3)
+    test_customer = cm.Customer.get(cm.Customer.customer_id==ok_customer3['customer_id'])
+
+    assert test_customer.email_address == ok_customer3['email_address']
```
codedragon: You are testing peewee here, not your code (well, your code for that add, but you are already testing that elsewhere). Otherwise, you would probably have discovered your search isn't doing what it is suppose to. :)

PR #50, opened 2019-01-30
Assignment 2 Update and Assignment 03 Partal needs update

timothyjpauley:
Hi, I went back and found where I went wrong on assignment 02. Assignment 03 still needs some work, but by next week, I should be all caught. Work has been busy and I have been learning JavaScript and Angular after work for my job.

Thanks, Tim


students/TimPauley/session02/code/charges_calc_update.py
```diff
@@ -0,0 +1,157 @@
...
+        logging.basicConfig(level=logging.CRITICAL)
+    if debug_level > 0:
+        logging.basicConfig(format=format_logger, level=logging.ERROR)
+        f_handler = logging.FileHandler(log_file)
+        f_handler.setLevel(logging.WARNING)
+        logging.getLogger().addHandler(f_handler)
+    if debug_level == 2:
+        logging.getLogger().setLevel(logging.WARNING)
+    if debug_level == 3:
+        logging.getLogger().setLevel(logging.DEBUG)
```
codedragon: This is great, keep in mind, if you aren't logging, you can just exit out immediately and not configure anything in this function.

students/TimPauley/session02/code/charges_calc_update.py
```diff
@@ -0,0 +1,157 @@
...
+    """
+    logging.debug("load_rentals_file")
+    logging.debug("Opens json file and readz it in")
+    with open(filename) as file:
+        try:
+            data = json.load(file)
+        except:
+            logging.error(error_msg)
+            logging.debug(i)
+    return data
```
codedragon: Remember if things go wrong, you should let your user know too, not just log it.

students/TimPauley/session02/code/charges_calc_update.py
```diff
@@ -0,0 +1,157 @@
...
+                    warning_price += f"{value['product_code']}"
+                    logging.warning(warning_price)
+                    logging.debug(value)
+                else:
+                    value['unit_cost'] = value['total_price'] / value[
+                        'units_rented']
+            except Exception as error_msg:
+                logging.error(error_msg)
+                logging.debug(value)
+                data_new = None
```
codedragon: You have some good error handling in this block. Keep in mind, you really want to limit what is in a try block to just the things you expect to go wrong. Try blocks should have max 3-4 lines in them.

students/TimPauley/session03/add_customer_test.py
```diff
@@ -0,0 +1,27 @@
...
+		):
+	'new_customer' = customer_id
+	,'name' = name
+	,'last_name' = last_name
+	,'home_address' = home_address
+	,'phone_number' = phone_number
+	,'email_address' = email_address
+	,'status' = status
+	,'credit_limit' = credit_limit
+new_customer.save()
```
codedragon: It looks like you are assigning strings to your variables. I don't think this is what you meant to do. Remember you have to use the create function, so the first line in this function will be something like 
```
new_customer = Customer.create(customer_id=customer_id,
    first_name=first_name,
    last_name=last_name,
    ...)
```
It worked out a bit strange that all of our variable names are the same as the column names for our model, but this is not unusual.

students/TimPauley/session03/basic_operations.py
```diff
@@ -0,0 +1,123 @@
...
+        customer_id = customer_id,
+        first_name = name,
+        last_name = last_name,
+        home_address = home_address,
+        phone_number = phone_number,
+        email_address = email_address,
+        status = status,
+        credit_limit = credit_limit)
+    new_customer.save()
+
```
codedragon: Lol, you did it correctly here. Please don't leave old code laying around, I have no way of knowing it is old code, when reviewing!

students/TimPauley/session03/basic_operations.py
```diff
@@ -0,0 +1,123 @@
...
+'''
+This function will return a 
+dictionary object with name, lastname, email address and phone number 
+of a customer or an empty dictionary object if no customer was found.
+'''
+def search_customer(customer_id):
+	for data_dict in data_source:
+    MyModel.create(**data_dict)
+
```
codedragon: Indentation doesn't look right here. Not to mention, I don't thing you want to create when you are searching. Copy and paste gone awry...

students/TimPauley/session03/create_customer.py
```diff
@@ -0,0 +1,31 @@
...
+
+logger.info('The next 3 lines of code are the only database specific code')
+
+database.create_tables([
+        Job,
+        Person,
+        PersonNumKey
+    ])
+
+database.close()
```
codedragon: I think you are confusing the assignment and the activity, just one table for the assignment.

students/TimPauley/session03/python_test.py
```diff
@@ -0,0 +1,28 @@
...
+		customer_id
+		,name
+		,last_name
+		,home_address
+		,email_address
+		,status
+		,credit_limit
+		)	
+		Customer.get(Customer.customer_id=customer_id)
+		assert(test_customer.email_address = email_address)
```
codedragon: I think indentation went awry again. Otherwise this test looks like it should work. Not sure if that is your actual information, but remember, these repos are public. For me, info like this is already available on the web, so it doesn't bother me, but want you to be aware of what you are doing, in case you feel differently.

students/TimPauley/session03/python_test_version2.py
```diff
@@ -0,0 +1,31 @@
...
+		,name
+		,last_name
+		,home_address
+		,phone_number
+		,email_address
+		,status
+		,credit_limit
+		)	
+		Customer.get(Customer.customer_id=customer_id)
+		assert(test_customer.email_address = email_address)
```
codedragon: This file looks very similar to the one above. Why have two?

PR #51, opened 2019-01-30
Lesson03 Pull - zconn

zconn:


students/ZackConnaughton/lessons/lesson03/assignment/customer_model.py
```diff
@@ -0,0 +1,44 @@
...
+
+def create_database():
+    if not os.path.isfile('customer.db'):
+        DATABASE.connect()
+        DATABASE.execute_sql('PRAGMA foreign_keys = ON;')
+
+        DATABASE.create_tables([
+            Customer
+            ])
+        DATABASE.close()
```
codedragon: Do you really want to do this if customer.db is NOT a file? Don't you want to create only if it is a file?

PR #53, opened 2019-02-05
Added assignment files under lesson04

dtingg:
I added functions for upload_csv and print_customers to basic_operations.py.  I also tried to add a dictionary comprehension under search_customer.

For test_basic_operations.py, I added some pytest fixtures.  My old tests still work and I included some new ones to test the csv upload.

I added a db.log file to capture the changes made to the database.  I'm also using a config.py file to write everything to test.db now.


students/DiannaTingg/lessons/lesson04/assignment/basic_operations.py
```diff
@@ -0,0 +1,140 @@
...
+                phone_number=phone_number,
+                email_address=email_address,
+                status=status,
+                credit_limit=credit_limit)
+            new_customer.save()
+            LOGGER.info("Successfully added Customer %s: %s %s to the database.", customer_id, first_name, last_name)
+    except pw.IntegrityError:
+        LOGGER.error("Error adding %s %s. Customer id %s already exists in the database.",
+                     first_name, last_name, customer_id)
+        raise pw.IntegrityError
```
codedragon: If you are raising the same error, you can just write raise.

students/DiannaTingg/lessons/lesson04/assignment/basic_operations.py
```diff
@@ -0,0 +1,140 @@
...
+def upload_csv(filename):
+    """
+    Adds customer data from a csv file.
+    """
+    with open(filename, newline="") as csvfile:
+        all_customers = csv.reader(csvfile)
+        headers = next(all_customers, None)  # Skip the header row
+
+        with cm.DATABASE.atomic():
+            cm.Customer.insert_many(all_customers, headers).execute()
```
codedragon: Woot! This is better than how I did it. I assume you got it working? :)

students/DiannaTingg/lessons/lesson04/assignment/basic_operations.py
```diff
@@ -0,0 +1,140 @@
...
+        customer_keys = ["first_name", "last_name", "email_address", "phone_number"]
+
+        customer_values = [current_customer.first_name, current_customer.last_name,
+                           current_customer.email_address, current_customer.phone_number]
+
+        customer_dict = dict(zip(customer_keys, customer_values))
+
+        LOGGER.info("Found customer %s: %s %s.", customer_id, current_customer.first_name, current_customer.last_name)
+
+        return customer_dict
```
codedragon: Nice. It is technically faster, as well as more in line with what we are learning, to do a dictionary comprehension, rather than the dict/zip combination. Also, this is too much for a try block. Think about which line is going to give you the does not exist exception, and only have that line in the try block.

students/DiannaTingg/lessons/lesson04/assignment/test_basic_operations.py
```diff
@@ -0,0 +1,164 @@
...
+
+def test_csv_update_customer_credit(csv_database):
+    bo.update_customer_credit("C000003", 100)
+
+    target_customer = cm.Customer.get(cm.Customer.customer_id == "C000003")
+
+    assert target_customer.credit_limit == 100
+
+    with pytest.raises(pw.DoesNotExist):
+        bo.update_customer_credit("200", 200)
```
codedragon: This test, and a few up above look like they should be split into two tests, one for being able to update the credit for a legit customer, and one that shows the error for updating a customer that doesn't exist. Tests should just test one thing. If you have more than one assert in a test, if the first fails, it does not run the rest of the test, so you lose information. Sometimes it makes sense to have more than one assert, but that is to test that one action has multiple outcomes.

PR #54, opened 2019-02-05
Lesson 04 Assignment

CodeChameleon:
Basic_operations:

- Added generator function to read csv file lines and output formatted dictionary for each line
- Added bulk insert method
- Moved insert customer validation to the bulk insert and pointed original add_customer to use bulk_insert
- Used dict comprehension for search_customer...in a future iteration I will update the "name" vs "first_name" property differences throughout the code so only one list is needed

Tests

- Existing tests continued to work
- Added test for bulk and CSV (10,000 records)
- Updated fixture to use the bulk insert method for setup, and used a generator for retrieving data, but that wasn't really necessary.


codedragon:
Looks great. 👍

PR #56, opened 2019-02-06
added lesson3 assignments

VioletLuo:
added files for lesson 3

### DIFF
--- a/students/ZidanLuo/lessons/lesson03/assignment/basic_operations.py
+++ b/students/ZidanLuo/lessons/lesson03/assignment/basic_operations.py
@@ -1,0 +1,73 @@
+import logging
+from customer_model import Customer
+import peewee
+
+
+def add_customer(customer_id, name, last_name, home_address, phone_number, email_address, status, credit_limit):
+    try:
+        credit_limit = float(credit_limit)
+    except ValueError as err:
+        logging.error(err)
+        raise
+    new_customer = Customer.create(
+        customer_id = customer_id,
+        first_name = name,
+        last_name = last_name,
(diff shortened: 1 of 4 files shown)


students/ZidanLuo/lessons/lesson03/assignment/basic_operations.py
```diff
@@ -0,0 +1,73 @@
...
+    their credit limit or raise a ValueError exception if the customer does 
+    not exist
+    :param:customer_id as as a string and credit_limit as a float
+    :return:None
+    '''
+    try:
+        customer = Customer.get(Customer.customer_id == customer_id)
+    except ValueError as err:
+        logging.error(err)
+        raise
```
codedragon: I think the error you should be looking for is a peewee DoesNotExist error. You should run the code without the try block, and see what error you get, and then look for that error. Also, in your add customer function, you are checking the value of credit limit to make sure it is a float, but here you are not.

students/ZidanLuo/lessons/lesson03/assignment/basic_operations.py
```diff
@@ -0,0 +1,73 @@
...
+from customer_model import Customer
+import peewee
+
+
+def add_customer(customer_id, name, last_name, home_address, phone_number, email_address, status, credit_limit):
+    try:
+        credit_limit = float(credit_limit)
+    except ValueError as err:
+        logging.error(err)
+        raise
```
codedragon: Excellent, limiting the number of lines in the try block is key to making troubleshooting easer. :)

PR #57, opened 2019-02-06
submitting hw4, what I have so far

kmonster462:
I am stuck trying to get the batch add functionality to work. I keep getting this error:

           ValueError: Missing value for "<CharField: Customers.last_name>".

and can't figure out where it's from. Once I get this function to work, I need to refactor the rest of my tests so they work.


codedragon:
Since I know you got this working, I'll wait for you to submit your new solution before I look at it again. :)

students/Kate_Koehn/Lesson05/database.py
```diff
@@ -0,0 +1,148 @@
...
+    try:
+        customers = db["customers"]
+        customers.insert_many(import_csv(customer_file))
+    except ImportError:
+        customer_errors += 1
+    try:
+        rentals = db["rentals"]
+        rentals.insert_many(import_csv(rentals_file))
+    except ImportError:
+        rental_errors += 1
```
codedragon: I think this doesn't actually return the number of errors per database, just whether there was an error or not. This is because you are doing insert many, and not doing a loop. What happens where there is an error with insert_many? Does it stop wherever it was? Is there anything in the ImportError that tells you how many errors there were?

students/Kate_Koehn/Lesson05/database.py
```diff
@@ -0,0 +1,148 @@
...
+    available_products = {}
+    for product_id in db.products.find():
+        product_dict = {"description": product_id["description"],
+                    "product_type": product_id["product_type"],
+                    "quantity_available": product_id["quantity_available"]}
+        if product_id["quantity_available"] != "0":
+            available_products[product_id["product_id"]] = product_dict
+            continue
+        else:
+            continue
```
codedragon: This could probably be made into a comprehension. :)

students/Kate_Koehn/Lesson05/database.py
```diff
@@ -0,0 +1,148 @@
...
+    for rental in db.rentals.find():
+        if rental["product_id"] == product_id:
+            customer_id = rental["user_id"]
+
+            customer_record = db.customers.find_one({"user_id": customer_id})
+
+            rental_users = {"name": customer_record["name"],
+                            "address": customer_record["address"],
+                            "phone_number": customer_record["phone_number"],
+                            "email": customer_record["email"]}
```
codedragon: This also could be simplified. The fact that your keys for your new dictionary are the same as the keys for the customer_record, could be used to your advantage. I'm not sure why you aren't just using customer_record. Is there extra stuff there?

PR #59, opened 2019-02-06
Submitting HW4

ktbzzz:


students/JaeKim/lesson04/assignment/basic_operations.py
```diff
@@ -0,0 +1,86 @@
...
+def search_customer(customer_id):
+    """
+    :param customer_id:
+    :param stage:
+    :return:
+    """
+    try:
+        results = Customer.get(Customer.customer_id == customer_id)
+    except Exception:
+        return False
```
codedragon: Remember to log and try to only catch the exception you expect (in this case I think it is peewee.DoesNotExist). If you aren't sure what you are looking for, temporarily get rid of the try block, purposely run bad data, and see what error you get.

students/JaeKim/lesson04/assignment/basic_operations.py
```diff
@@ -0,0 +1,86 @@
...
+    """
+    :param customer_id:
+    :param credit_limit:
+    :param stage:
+    :return:
+    """
+
+    customer = Customer.get(Customer.customer_id == customer_id)
+    update = customer.update(credit_limit=credit_limit)
+    update.execute()
```
codedragon: Notice up above you are converting to a float and checking for errors, but here you are not. Consistency is key when working with data.

students/JaeKim/lesson04/assignment/customers_model.py
```diff
@@ -0,0 +1,70 @@
...
+    LOGGER.info('Must be a unique identifier for each person')
+
+    customer_id = pw.CharField(primary_key=True, max_length=30)
+    first_name = pw.CharField(max_length=30)
+    last_name = pw.CharField(max_length=30)
+    home_address = pw.CharField(max_length=100)
+    phone_number = pw.CharField(max_length=20)
+    email_address = pw.CharField(max_length=100)
+    status = pw.BooleanField()
+    credit_limit = pw.FloatField()
```
codedragon: This is a good idea, separating test and prod, but here is not where to do it. This means you have to be diligent to keep the two classes exactly the same. But, you can use the same class, you just need to use a different database. Feel free to look at the solution in my student directory to see how this can be done.

students/JaeKim/lesson04/assignment/tests/test_basic_operations.py
```diff
@@ -0,0 +1,202 @@
...
+                # customer_id,name,last_name,home_address,phone_number,email_address,status,credit_limit
+                dict_user['customer_id'] = row[0]
+                dict_user['name'] = row[1]
+                dict_user['lastname'] = row[2]
+                dict_user['home_address'] = row[3]
+                dict_user['phone_number'] = row[4]
+                dict_user['email_address'] = row[5]
+                dict_user['status'] = row[6]
+                dict_user['credit_limit'] = row[7]
+                data.append(dict_user)
```
codedragon: This looks like an excellent place for a dictionary comprehension. :) Also, this function should be in basic_operations, because tests aren't the only place you might want to add customers from a csv file. Then, finally, you might want to think about making this function a generator. Then you don't have to do the iter call down below.

students/JaeKim/lesson04/assignment/tests/test_basic_operations.py
```diff
@@ -0,0 +1,202 @@
...
+
+@pytest.fixture(scope="function")
+def setup_test_db():
+    cc.db.drop_tables([
+        Customer
+    ])
+    cc.db.create_tables([
+        Customer
+    ])
+    cc.db.close()
```
codedragon: You actually shouldn't close the database here. Only close the database connection when you are completely done.

students/JaeKim/lesson04/assignment/tests/test_basic_operations.py
```diff
@@ -0,0 +1,202 @@
...
+    valid_customer = search_customer(OK_CUSTOMER['customer_id'])
+
+    assert valid_customer.first_name == OK_CUSTOMER['name']
+    assert valid_customer.customer_id == OK_CUSTOMER['customer_id']
+    assert valid_customer.last_name == OK_CUSTOMER['lastname']
+    assert valid_customer.home_address == OK_CUSTOMER['home_address']
+    assert valid_customer.phone_number == OK_CUSTOMER['phone_number']
+    assert valid_customer.email_address == OK_CUSTOMER['email_address']
+    assert valid_customer.status == OK_CUSTOMER['status']
+    assert valid_customer.credit_limit == OK_CUSTOMER['credit_limit']
```
codedragon: If you are asserting a delete, you can assume the customer is there, since you tested adding above. Just add, and then make sure you can delete.

PR #63, opened 2019-02-11
Bulk add customers

rushinstuffin:
Bulk add customers function created.

### DIFF
--- a/students/KyleBarry/lesson03/basic_operations.py
+++ b/students/KyleBarry/lesson03/basic_operations.py
@@ -1,10 +1,38 @@
-from customer_model import Customer
 import peewee as pw
 import logging
+import csv
+from customer_model import Customer
 
-logging.basicConfig(level=logging.INFO)
-logger = logging.getLogger(__name__)
+log_format = "%(asctime)s %(filename)s:%(lineno)-4d %(levelname)s %(message)s"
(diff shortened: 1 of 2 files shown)


students/KyleBarry/lesson03/test_customer.py
```diff
@@ -56,7 +72,7 @@ def test_list_active_customers():
 
     result = list_active_customers()
     logging.info(result)
-    assert result == 1
+    assert result == 8
```
codedragon: Your tests are dependent on order, which is difficult to maintain, and means it is impossible to run tests in isolation. Consider using fixtures. https://docs.pytest.org/en/latest/fixture.html

students/KyleBarry/lesson03/basic_operations.py
```diff
@@ -17,7 +45,7 @@ def add_customer(customer_id, first_name, last_name, home_address, phone_number,
                                status=status,
                                credit_limit=credit_limit)
         cust.save()
-        logger.info("Customersuccessfully added!")
+        logging.info("Customer successfully added!")
 
     except Exception as e:
         logging.info(f"Could not add {first_name} to db")
```
codedragon: It is best to catch the specific exception that you are expecting, like you did above. Removing the try/except block and causing the exception to happen on purpose is a good way to figure out what exception you are looking for. Also you should log the exception itself. ```logging.error(e)```

PR #64, opened 2019-02-11
Added assignment files under lesson05

dtingg:
My database.py script seems to work and everything in test_database.py passes, but I had a few questions:

I couldn't get most of the functions in database.py to work unless I passed "db" in everywhere.  Is there a better way to do that?

I also wasn't sure the best way to set up the pytest fixture for MongoDB.  What I have seems to work though.

I couldn't get my _add_bulk_data function to throw any errors (even with duplicate data).  Does the except part of my function look ok?  Is there a better way to return the number of errors?

Thanks so much Maria!


students/DiannaTingg/lessons/lesson05/assignment/database.py
```diff
@@ -0,0 +1,185 @@
...
+        headers = next(csv_data, None)  # Save the first line as the headers
+
+        if headers[0].startswith("ï»¿"):  # Check for weird formatting
+            headers[0] = headers[0][3:]
+
+        for row in csv_data:
+            row_dict = {}
+
+            for index, column in enumerate(headers):
+                row_dict[column] = row[index]
```
codedragon: Nice place for a dictionary comprehension.
```
row_dict = {column: row[index] for index, column in enumerate(headers)}
```
:)

students/DiannaTingg/lessons/lesson05/assignment/database.py
```diff
@@ -0,0 +1,185 @@
...
+    available_products = {}
+
+    for product in db.products.find():
+        if int(product["quantity_available"]) > 0:
+
+            product_dict = {"description": product["description"],
+                            "product_type": product["product_type"],
+                            "quantity_available": product["quantity_available"]}
+
+            available_products[product["product_id"]] = product_dict
```
codedragon: product and product_dict look very similar. Are the keys/values in product that you don't want in product_dict? Is that why you aren't just making a copy? If it is just one key, you could make a copy and delete that key. Or do a comprehension (where I'm putting it directly into the available_products dictionary)
```
available_products[product["product_id"]] = {key: value for key, value in product.items() if key is not "quantity_available"}
```
It is a pretty long line, so may be better to save it first, then enter into dict. 
This can be done in show_rentals, as well, I think.

students/DiannaTingg/lessons/lesson05/assignment/test_database.py
```diff
@@ -0,0 +1,71 @@
...
+    :return:
+    """
+    mongo = d.MongoDBConnection()
+
+    with mongo:
+        db = mongo.connection.media
+
+        yield db
+
+        d.clear_data(db)
```
codedragon: This seems like a reasonable fixture. :)

PR #65, opened 2019-02-12
Lesson 5 assignment

vanulor:
New assignment, using fixtures, a set of CSV files, and MongoDB.  I am including both the code and the tests.

### DIFF
--- a/students/BrianNisonger/Lesson5/assignment/database.py
+++ b/students/BrianNisonger/Lesson5/assignment/database.py
@@ -1,0 +1,41 @@
+import os
+import csv
+import logging
+from pymongo import MongoClient
+
+class MongoDBConnection(object):
+    def __init__(self, host='127.0.0.1', port=27017):
+        self.host = host
+        self.port = port
+        self.connection = None
+
+    def __enter__(self):
+        self.connection = MongoClient(self.host, self.port)
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.connection.close()
+
+def import_data(directory,product_file,customer_file,rental_file):
+    mongo = MongoDBConnection()
+    with mongo:
+        db = mongo.connection.furniture
+        files=[product_file,customer_file,rental_file]
+        added_count_list=[]
+        error_count_list=[]
+        for file in files:
+            with open(os.path.join(directory,file)) as csv_file:
+                csv_dict = csv.DictReader(csv_file, delimiter=',')
+                collection = db[file.replace(".csv","")]
+                result = collection.insert_many(csv_dict)
+                added_count=result("insertedCount")
+                error_count=0
+                added_count_list.append(added_count)
+                error_count_list.append(error_count)
+    return added_count_list,error_count_list
+    
+    def show_available_products():
+        pass
+    
+    def show_rentals():
+        pass
--- a/students/BrianNisonger/Lesson5/assignment/database.py
+++ b/students/BrianNisonger/Lesson5/assignment/database.py
@@ -27,12 +27,14 @@ def import_data(directory,product_file,customer_file,rental_file):
             with open(os.path.join(directory,file)) as csv_file:
                 csv_dict = csv.DictReader(csv_file, delimiter=',')
                 collection = db[file.replace(".csv","")]
-                result = collection.insert_many(csv_dict)
-                added_count=result("insertedCount")
-                error_count=0
-                added_count_list.append(added_count)
-                error_count_list.append(error_count)
-    return added_count_list,error_count_list
+                try:
+                    result = collection.insert_many(csv_dict)
+                    added_count_list.append(len(result.inserted_ids))
+                    error_count_list.append(0)
+                except BulkWriteError as bwe:
+                    error_count_list.append(1)
+                    logging.error(bwe)
+    return (tuple(added_count_list),tuple(error_count_list))
     
     def show_available_products():
         pass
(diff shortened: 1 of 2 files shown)


students/BrianNisonger/Lesson5/assignment/database.py
```diff
@@ -0,0 +1,108 @@
...
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.connection.close()
+
+
+def calculate_availability(product_list, rental_list):
+    for product in product_list:
+        count = 0
+        for rental in rental_list:
+            if rental["product_id"] == product["product_id"]:
+                count += 1
```
codedragon: A doc string would be useful here, and everywhere. :) You can use sum with a comprehension:
```
count = sum(1 for rental in rental_list if rental["product_id"] == product["product_id"])
```

students/BrianNisonger/Lesson5/assignment/database.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+def make_product_dict(prod_available_list):
+    fields = ["description", "product_type", "quantity_available"]
+    product_values_dict = {}
+    product_dict = {}
+    for product in prod_available_list:
+        product_values_dict = {}
+        for field in fields:
+            product_values_dict[field] = product[field]
```
codedragon: This for loop is a prefect candidate for a comprehension.

students/BrianNisonger/Lesson5/assignment/database.py
```diff
@@ -0,0 +1,108 @@
...
+        if product["quantity_available"] > 0:
+            product_dict[product["product_id"]] = product_values_dict
+    return product_dict
+
+
+def return_user_ids(rental_list):
+    return [rental["user_id"] for rental in rental_list]
```
codedragon: Nice comprehension.

students/BrianNisonger/Lesson5/assignment/database.py
```diff
@@ -0,0 +1,108 @@
...
+
+def make_customer_dict(user_list):
+    fields = ["name", "address", "phone_number", "email"]
+    customer_values_dict = {}
+    customer_dict = {}
+    for customer in user_list:
+        customer_values_dict = {}
+        for field in fields:
+            customer_values_dict[field] = customer[field]
+            customer_dict[customer["user_id"]] = customer_values_dict
```
codedragon: I think you don't want this line:
```
customer_dict[customer["user_id"]] = customer_values_dict
```
inside your second for loop (which is also a candidate for a comprehension), because you are adding the customer_values_dict every field instead of after the dictionary is fully populated.

students/BrianNisonger/Lesson5/assignment/database.py
```diff
@@ -0,0 +1,108 @@
...
+            with open(os.path.join(directory, file)) as csv_file:
+                csv_dict = csv.DictReader(csv_file, delimiter=',')
+                collection = db[file.replace(".csv", "")]
+                try:
+                    result = collection.insert_many(csv_dict)
+                    added_count_list.append(len(result.inserted_ids))
+                    error_count_list.append(0)
+                except BulkWriteError as bwe:
+                    error_count_list.append(1)
+                    logging.error(bwe)
```
codedragon: BulkWriteError will quit if it gets an error, so doing it this way, you would have to re-try your insert_many every time an error occurs.

students/BrianNisonger/Lesson5/assignment/test_database.py
```diff
@@ -0,0 +1,177 @@
...
+        }
+    }
+    product_dict_list = product_list = [{
+        "product_id": "prd1001",
+        "quantity_available": 0,
+        "description": "blah",
+        "product_type": "yes",
+        "garbage_field": "yuck"
+    }]
+    assert make_product_dict(product_dict_list) == {}
```
codedragon: These should be two separate tests. They are testing separate concepts (what happens with quantity available is more than one vs 0). Even better, use parameterize. https://docs.pytest.org/en/latest/parametrize.html

students/BrianNisonger/Lesson5/assignment/test_database.py
```diff
@@ -0,0 +1,177 @@
...
+        self.host = host
+        self.port = port
+        self.connection = None
+
+    def __enter__(self):
+        self.connection = MongoClient(self.host, self.port)
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.connection.close()
```
codedragon: This looks the same as your mongo connection class in your other file. Just import that one, you don't have to create it again.

students/BrianNisonger/Lesson5/assignment/test_database.py
```diff
@@ -0,0 +1,177 @@
...
+        db.rental_data.drop()
+        import_data("dat", "product_data.csv", "customer_data.csv",
+                    "rental_data.csv")
+
+        def cleanup():
+            db.product_data.drop()
+            db.customer_data.drop()
+            db.rental_data.drop()
+
+        request.addfinalizer(cleanup)
```
codedragon: Great fixture. :)

students/BrianNisonger/Lesson5/assignment/test_database.py
```diff
@@ -0,0 +1,177 @@
...
+            'product_type': 'livingroom',
+            'quantity_available': 2
+        },
+        'prd1003': {
+            'description': 'Hobart Stand Mixer',
+            'product_type': 'kitchen',
+            'quantity_available': 10
+        }
+    }
+    return product_dict
```
codedragon: It is a little confusing calling this show_available_products. Why is it a function?

PR #66, opened 2019-02-13
Assignment 5 Mongo DB

CodeChameleon:
Working with MongoDB was interesting. I've wanted to explore MongoDB for a while. It was interesting to find out that there is no real "join" between collections (at least from what I found). You pull all of the documents and process them locally. I found the  "$gt" and "$in" operators to allow MongoDB do some of the processing. I will explore those options more.

I did use a config.py file this time to store the names of the production and test databases so the tests will always alter the test instance and not prod. I added keyword parameters so that connection and database_name values can be passed to each function.


students/DrewSmith/lessons/lesson05/assignment/database.py
```diff
@@ -0,0 +1,133 @@
...
+                                                        "email"))
+            records = [record for index, record in enumerate(csv_file) if index > 0]
+            counts[1] = len(records)
+            database["customers"].insert_many(records)
+
+        with open(directory / rentals_file, 'r') as file:
+            csv_file = csv.DictReader(file, fieldnames=("product_id", "user_id"))
+            records = [record for index, record in enumerate(csv_file) if index > 0]
+            counts[2] = len(records)
+            database["rentals"].insert_many(records)
```
codedragon: Nice comprehensions. :)
This is a lot for one function, I would think about breaking some of this out as separate functions. Each of your context manager blocks look pretty similar, I bet you could take advantage of that. Also it seems you only expect to get one type of error, and only for one file. That seems optimistic. ;-)

students/DrewSmith/lessons/lesson05/assignment/database.py
```diff
@@ -0,0 +1,133 @@
...
+    database_name = database_name or config.DATABASE_NAME
+    with connect:
+        database = connect.connection.get_database(name=database_name)
+        products = database["products"]
+        result = dict()
+        for record in products.find({"quantity_available": {"$gt": 0}}):
+            result[record["product_id"]] = {
+                "description": record["description"],
+                "product_type": record["product_type"],
+                "quantity_available": record["quantity_available"]
```
codedragon: This dictionary seems very similar to the one you are given back from MongoDb. Are you doing it this way because you don't want to include something? You could make a copy, and delete the key you don't want, or do a comprehension:
```
resut[record["product_id"]] = {key: value for key, value in record.items() if key is not "one I don't want"}
```

students/DrewSmith/lessons/lesson05/assignment/database.py
```diff
@@ -0,0 +1,133 @@
...
+    :param directory_name: directory where data files are stored
+    :param product_file: file name with product data
+    :param customer_file: file name with customer data
+    :param rentals_file: file name with rentals data
+    :param connection: DB connection for data inserts
+    :param database_name: Database to use for the query,
+        or None to use production default
+    '''
+    connection = connection or MongoDBConnection()
+    database_name = database_name or config.DATABASE_NAME
```
codedragon: Great way of handling using separate databases. :)

students/DrewSmith/lessons/lesson05/assignment/tests/test_database.py
```diff
@@ -0,0 +1,68 @@
...
+def setup_teardown():
+    ''' Fixture to execute before and after tests '''
+    mongo = assignment.database.MongoDBConnection()
+    with mongo:
+        db = mongo.connection.get_database(name=config.TEST_DATABASE_NAME)
+        db.drop_collection("customers")
+        db.drop_collection("products")
+        db.drop_collection("rentals")
+    
+    yield
```
codedragon: Great use of fixture

PR #67, opened 2019-02-13
Lesson05 Pull request

zconn:
added csv files as well, I added in one that is supposed to be a problem and not load so I could see what would happen if there is an error.


students/ZackConnaughton/lessons/lesson05/assignment/database.py
```diff
@@ -0,0 +1,140 @@
...
+def print_mdb_collection(collection_name):
+    for doc in collection_name.find({}):
+        print(doc)
+
+def csv_list(file):
+    return_list = []
+    with open('sample_csv_files/' + file, encoding='utf-8-sig') as csv_file:
+        file_input = csv.DictReader(csv_file)
+        for row in file_input:
+            return_list.append(row)
```
codedragon: Great place for a comprehension! 
```return_list = [row for row in file_input]```
but, if that is what you want to do, you can do ```file_input = list(csv.DictReader(csv_file))```
However, I don't think that is what you really want to do. If you are going to iterate through it, you should just leave it as an iterator. Much better memory management.

File: students/ZackConnaughton/lessons/lesson05/assignment/database.py [hunk 2]
```diff
@@ -0,0 +1,140 @@
...
+            except:
+                rental_list_errors += 1
+
+        import_data_added = (product_list_count - product_list_errors,
+                             customer_list_count - customer_list_errors,
+                             rental_list_count - rental_list_errors)
+        import_data_errors = (product_list_errors,
+                              customer_list_errors,
+                              rental_list_errors)
+
```
codedragon: This function is pretty big, and it looks like you are doing basically the same thing several times. I would make a function so you don't have the same code written several times.

students/ZackConnaughton/lessons/lesson05/assignment/test_database.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+@pytest.fixture(scope='function', autouse=True)
+def setup_database(request):
+    test_entry, test_errors = database.import_data('main_mongo', 'products.csv', 'customers.csv', 'rentals.csv')
+    def remove_database():
+        database.drop_database_data('products')
+        database.drop_database_data('customers')
+        database.drop_database_data('rentals')
+    request.addfinalizer(remove_database)
```
codedragon: Nice use of fixtures!

students/ZackConnaughton/lessons/lesson05/assignment/test_database.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def test_show_rentals():
+    """
+    returns dict with user id, name, address, phone number, email
+    """
+    prod2_rentals = database.show_rentals('prd002')
+    assert (item for item in prod2_rentals if item['name'] == 'Shirlene Harris')
+    prod00X_rentals = database.show_rentals('prodX')
+    assert prod00X_rentals == []
```
codedragon: This should be two separate tests, or one test using parameterize. This way, if the first one fails, the second one will still run, and you will have more complete information about what works and what doesn't. https://docs.pytest.org/en/latest/parametrize.html

PR #69, opened 2019-02-13
Lesson 05 Assignment

Zach-Cooper:
Still working a bit on my tests for the database. On my main database.py you will see two parts. The first part was my initial attempt to load MongoDB and that worked successfully but I ran into trouble later on in the assignment because I was importing the csv headers as indeces and it was tough using count for that. The second part is cleaner code but now I am having trouble establishing the DB and populating it. I think I am close with this but cant figure out the difference between my frist attempt and second attempt. Any thoughts???


students/Zach Cooper/lesson05/database.py
```diff
@@ -0,0 +1,209 @@
...
+    products.insert_many(csv_import(products_file_path))
+
+    # Rentals Collection in db
+    rentals = db["rentals"]
+    rentals_file_path = os.path.join(directory_name, rentals_file)
+    rentals.insert_many(csv_import(rentals_file_path))
+
+    record_count = (db.customers.count_documents({}), db.products.count_documents({}), db.rentals.count_documents({}))
+
+    return record_count
```
codedragon: Note you are suppose to look for errors too.

students/Zach Cooper/lesson04/basic_operations_refactored.py
```diff
@@ -0,0 +1,150 @@
...
+    find_customer_dict = {}
+    LOGGER.info("Search for customer name in database...")
+    try:
+        find_customer = Customer.get(Customer.customer_id == customer_id)
+        if find_customer:
+
+            find_customer_dict = {"first_name": find_customer.first_name,
+                                  "last_name": find_customer.last_name,
+                                  "email_address": find_customer.email_address,
+                                  "phone_number": find_customer.phone_number}
```
codedragon: Here is a place for a comprehension, but you have to know about getattr. Getattr gets a attribute from an object, so find_customer.first_name is equivalent to getattr(find_customer, 'first_name'), which is easier to substitute a variable in.
```
my_attrs = ['first_name', 'last_name', 'email_address', 'phone_number']
find_customer_dict = {cust_attr: getattr(find_customer, cust_attr) for cust_attr in my_attrs}
```

PR #72, opened 2019-02-18
Added assignment files under lesson06

dtingg:
I refactored the code in good_perf.py and I created one test.  All of my notes are in notes.txt.  

I got the function down to 6.8 seconds, but not sure how to make it any faster.

Thank you!

### DIFF
--- a/students/DiannaTingg/lessons/lesson06/assignment/good_perf.py
+++ b/students/DiannaTingg/lessons/lesson06/assignment/good_perf.py
@@ -1,0 +1,59 @@
+"""
+Lesson 06 Assignment
+"""
+
+import datetime
+import csv
+# from timeit import timeit
+# from line_profiler import LineProfiler
+
+
+def analyze(filename):
+    """
+    Returns certain statistics for a file.
+    :param filename: csv file
+    :return: Start date, end date, frequency of years 2013-2018, frequency of "ao"
+    """
+    start = datetime.datetime.now()
+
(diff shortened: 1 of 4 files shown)


students/DiannaTingg/lessons/lesson06/assignment/good_perf.py
```diff
@@ -0,0 +1,59 @@
...
+
+        next(reader, None)  # Skip the header row.
+
+        year_count = {}
+
+        found = 0
+
+        for row in reader:
+            if "2012" < row[5][-4:] < "2019":
+                year_count[row[5][-4:]] = year_count.get(row[5][-4:], 0) + 1
```
codedragon: I've been thinking about this. Here is how to do it with setdefault (amusingly pretty much exactly the same way as using get):
```
year_count[row[5][-4:]] = year_count.setdefault(row[5][-4:], 0) + 1
```
Which I'm told is actually faster than using the get method I showed you. And even faster, is using defaultdict, but you do have to use a lambda
```
from collections import defaultdict
year_count = defaultdict(lambda: 0)
year_count[row[5][-4:]] += 1
```
Finally, you could use a comprehension to populate the dictionary ahead of time:
```
year_count = {str(year): 0 for year in range(2013, 2019)}
```
I actually suspect that is slightly less efficient, but maybe more readable, and for the amount of efficiency gained, that may be worth it.

PR #73, opened 2019-02-18
Assignment 6

vanulor:
I was able to shave 2 full seconds off the runtimes.  Here are my results:

Poor:                                                                                                                                                                                            {'2013': 5911, '2014': 5854, '2015': 5994, '2016': 5762, '2017': 11600, '2018': 0}                                                                                                               'ao' was found 63395 times                                                                                                                                                                       11.2488242                                                                                                                                                                                       Good:                                                                                                                                                                                            {'2013': 5911, '2014': 5854, '2015': 5994, '2016': 5762, '2017': 5789, '2018': 5811}                                                                                                             'ao' was found 63395 times                                                                                                                                                                       9.1343994
--

I removed the extra opening of the file by re-seeking back to the beginning.  I swapped out for loops and if statements with getters and comprehensions.  I used getters when I could because I know they run faster than comprehensions.   Instead of a costly creation of a dictionary and nested if I used a counter and then filtered it using a dictionary comprehension.  I fixed the error which was adding 2018 numbers to 2017 in my version.  The final portion uses a length of a list which only contains the thing I am looking for.  This shaved about a 1/10th of a second off my results as to just using a for loop and a if statement with the getter so I went with this slightly more complicated method. I used the timeit to compare each change to the original code to ensure that each change actually saved me time. I created a separate file which ran the original module and the optimized module and then printed out the time it took for each.  This allowed me to test different code and see what worked.  This is my final optimized code.


students/BrianNisonger/Lesson6/good_perf.py
```diff
@@ -0,0 +1,41 @@
...
+def analyze(filename):
+"""
+Analyzes a file and produces a count of years and the number of times "ao" was found
+"""
+    start = datetime.datetime.now()
+    with open(filename) as csvfile:
+        reader = csv.reader(csvfile, delimiter=',', quotechar='"')
+        new_ones = (row[5] for row in reader)
+        getter_of_years = (y[6:] for y in new_ones)
+        year_count = collections.Counter(getter_of_years)
```
codedragon: Nice find! I think might be able save more, by doing more in your comprehension:
```
getter_of_years = (row[5][6:] for row in reader if row[5][6:] > "2012" and row[5][6:] < "2019")
```
And then, I think you have your dict without having to do any filtering.

students/BrianNisonger/Lesson6/good_perf.py
```diff
@@ -0,0 +1,41 @@
...
+        filtered_year_dict = {
+            year: year_count[year]
+            for year in ["2013", "2014", "2015", "2016", "2017", "2018"]
+        }
+        print(filtered_year_dict)
+        csvfile.seek(0)
+        found = 0
+        found_getter = (line[6] for line in reader)
+        found = len([1 for x in found_getter if 'ao' in x])
```
codedragon: Again, I think it will be even faster if you combine:
```
found = len([1 for line in reader if 'ao' in line[6]])
```

PR #75, opened 2019-02-19
lesson06 assignment with good_perf py module and notes

habtamua:


students/HABTAMU/lesson06/assignment/good_perf.py
```diff
@@ -0,0 +1,37 @@
...
+        new_ones = []
+        found = 0
+        for row in reader:
+            if (row[5][6:] > '2012') and (row[5][6:] < '2019'):
+                year_count[row[5][6:]] += 1
+
+            if "ao" in row[6]:
+                found +=1
+
+        year_count = dict(year_count)
```
codedragon: You don't need this line, and it probably is slowing you down a bit. Defaultdict is a dictionary, even though it has the weird signature.

habtamua: the reason i added that was to remove that weird object text ... 
this is what am talking about... `defaultdict(<function analyze.<locals>.<lambda> at 0x109d3fa60>, {'2015': 5994, '2013': 5911, '2014': 5854, '2016': 5762, '2017': 5789, '2018': 5811})` when i use dict(year) it will not show that again.

PR #76, opened 2019-02-19
Assignment 6

CodeChameleon:
Poor Performance Timeit: **9.248711s**
Good Performance Timeit: **2.04s**
Final with datetime parsing: **16.5s**

SUMMARY: Biggest performance increases were from 
	1. Removing duplicate looping (original was looping over data 2.5 times)
	2. Switching from CSV library parsing to just using file.readline() to get the line string and slicing the values needed using .rfind(',') to get last comma index
	
Detailed notes on what I tried (both good and bad) are in the Perf_notes.txt file.

Per your Slack note I tried datetime parsing. Was able to go from 71 seconds using dateutils.parse down to 16.5 using datetime.strptime().
Datetime parsing code is commented out in my code.

The poor_perf.py code had this possible bug, but I coded as though that was a feature. (2018 gets counted into 2017).
Fixing would speed the code up a bit since I could just do a straight dictionary lookup
if new[0][6:] == '2018':
	year_count["2017"] += 1

### DIFF
--- a/students/DrewSmith/lessons/lesson06/assignment/good_perf.py
+++ b/students/DrewSmith/lessons/lesson06/assignment/good_perf.py
@@ -1,0 +1,70 @@
+"""
+good performing, perfectly written module :)
+
+"""
+
+
+import datetime
+from dateutil.parser import parse
+from datetime import datetime as dt
+
+def analyze(filename):
+    '''
+    Reads and analyzes data from the given file
+
+    :param filename: file path of file to analyze
+    '''
+    start = datetime.datetime.now()
+    
+    year_count = {
+            "2013": 0,
+            "2014": 0,
+            "2015": 0,
+            "2016": 0,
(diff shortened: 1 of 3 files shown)


students/DrewSmith/lessons/lesson06/assignment/good_perf.py
```diff
@@ -0,0 +1,70 @@
...
+            "2017": 0,
+            "2018": 0
+    }
+    found = 0
+    with open(filename) as csvfile:      
+        row = csvfile.readline()
+        # skip the first record
+        row = csvfile.readline()
+        while row:
+            comma_index = row.rfind(",")
```
codedragon: Excellent. You are the first person to question using the csv library, and it paid off. If you were really going to change this in production code, you would want to do some research into what you are losing by not using the csv library, and if the tradeoffs are worth it.

PR #77, opened 2019-02-19
Assignment04 and 05 completed

Misseswaddles:


students/cfarris/lessons/lesson05/assignment/database.py
```diff
@@ -0,0 +1,167 @@
...
+        first_line = []
+        coll_of_dicts = []
+        for row in csv_reader:
+            if header:
+                header = False
+                first_line = [value.strip() for value in row]
+            else:
+                new_dict = {}
+                str_row = []
+                str_row = [value.strip() for value in row]
```
codedragon: You don't need to initialize str_row with an empty list if you are doing a list comprehension. Nice comprehension.

students/cfarris/lessons/lesson05/assignment/database.py
```diff
@@ -0,0 +1,167 @@
...
+                for i in range(len(str_row)):
+                    new_dict[first_line[i]] = str_row[i]
```
codedragon: You could get super fancy, and do a dictionary comprehension here.
```
new_dict = {first_line[index]: value.strip() for index, value in enumerate(row)}
```
 I think that works, but didn't test it...

students/cfarris/lessons/lesson05/assignment/test_database.py
```diff
@@ -0,0 +1,140 @@
...
+        self.host = host
+        self.port = port
+        self.connection = None
+
+    def __enter__(self):
+        self.connection = MongoClient(self.host, self.port)
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.connection.close()
```
codedragon: You can just import this, you should never be typing stuff like this out more than once...

students/cfarris/lessons/lesson05/assignment/test_database.py
```diff
@@ -0,0 +1,140 @@
...
+                                  'quantity_available': '20'},
+                       'prd006': {'description': 'Portable heater',
+                                  'product_type': 'bathroom',
+                                  'quantity_available': '14'},
+                       'prd008': {'description': 'Smart microwave',
+                                  'product_type': 'kitchen',
+                                  'quantity_available': '30'},
+                       'prd010': {'description': '60-inch TV',
+                                  'product_type': 'livingroom',
+                                  'quantity_available': '3'}}
```
codedragon: This is starting to get unwieldy. Think about how the information can be available for both the tests and the code you are testing.

PR #78, opened 2019-02-20
Lesson 06 Assignment

zconn:
Assignment for Lesson 06


students/ZackConnaughton/lessons/lesson07/assignment/database.py
```diff
@@ -0,0 +1,234 @@
...
+def show_available_products():
+    mongo = MongoDBConnection()
+
+    with mongo:
+        database = mongo.connection.media
+
+        output_dict = list(database.products.find({'quantity_available': {'$gt':'0'}}, {"_id": 0}))
+
+        print(output_dict)
+        return(output_dict)
```
codedragon: Once you have your output_dict, you should close out of the context manager:
```
with mongo:
    database = mongo.connection.media
    output_dict = list(database.products.find({'quantity_available': {'$gt':'0'}}, {"_id": 0}))
print(output_dict)
return output_dict
```
True for other functions as well. The unindent closes the connection.

students/ZackConnaughton/lessons/lesson07/assignment/database.py
```diff
@@ -0,0 +1,234 @@
...
+    for doc in collection_name.find({}):
+        print(doc)
+
+def csv_list(file):
+    return_list = []
+    with open('sample_csv_files/' + file, encoding='utf-8-sig') as csv_file:
+        file_input = csv.DictReader(csv_file)
+        for row in file_input:
+            return_list.append(row)
+    return return_list
```
codedragon: return_list looks like it could be a comprehension pretty easily.

students/ZackConnaughton/lessons/lesson07/assignment/database.py
```diff
@@ -0,0 +1,234 @@
...
+
+    data_entry_queue.join()
+
+    output_raw = []
+    for item in range(len(threads)):
+        output_raw.append(data_entry_queue.get())
+
+    print(output_raw)
+    output = format_output(output_raw)
+    return output
```
codedragon: How well did your threads work? It looks pretty much how I would expect, but it depends on whether mongodb can thread if it saves any time. If it were network connections, there would probably be a more obvious gain. Nice job.

students/ZackConnaughton/lessons/lesson07/assignment/test_database.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def test_show_rentals():
+    """
+    returns dict with user id, name, address, phone number, email
+    """
+    prod2_rentals = database.show_rentals('prd002')
+    assert (item for item in prod2_rentals if item['name'] == 'Shirlene Harris')
+    prod00X_rentals = database.show_rentals('prodX')
+    assert prod00X_rentals == []
```
codedragon: Nice tests. This last one should probably be using parameterize, since it is basically the same function being tested with different inputs/outputs.

students/ZackConnaughton/lessons/lesson08/assignment/inventory.py
```diff
@@ -0,0 +1,28 @@
...
+        reader = csv.reader(file)
+        with open(output_file, "a+") as output_file:
+            for row in reader:
+                if row[0] == customer:
+                    output_file.write(",".join(row))
+                    output_file.write("\n")
+
+
+def single_customer(customer, output_file):
+    return partial(rental_items, customer=customer, output_file=output_file)
```
codedragon: This is close, but not quite what was asked for. There is no closure here, which was part of the directions. Notice that you are writing some of the same code in rental_items as you did in add_furniture. The idea was to change the signature of add_furniture, using partial, and then use that in the function that you return. So, there will be two functions total, not three. It is not efficient,  so you wouldn't do it in real life, but it teaches you to use partial and closures. :)

students/ZackConnaughton/lessons/lesson08/assignment/test_inventory.py
```diff
@@ -0,0 +1,51 @@
...
+    assert line_count == 2
+    remove_file(elisa_invoice_csv)
+
+def test_single_customer_bad(setup_invoice_file):
+    non_customer_csv = 'non_customer.csv'
+    create_invoice = single_customer("Non Customer", non_customer_csv)
+    create_invoice(INVOICE_FILE)
+    line_count = sum(1 for line in open(non_customer_csv))
+    assert line_count == 0
+    remove_file(non_customer_csv)
```
codedragon: Your tests should still work when you change your single_customer function, so it will be easy to see if you succeeded in refactoring. Tests are so useful. :)

PR #81, opened 2019-02-21
Added lesson06 Assignment

Zach-Cooper:
I created to different .py files to compare results for this assignment. good_perf converts datetime strings to date objects which we did in office hours but output time is just too long. alt_good_perf keeps the dates as strings and is much quicker. Where could I change code in good_perf to possibly make it faster? Not sure there is much more I can do since the majority of the time it takes is directed towards converting all the datetimes.
Thanks.

### DIFF
--- a/students/Zach Cooper/lesson06/Description of Changes Made to good_perf.py
+++ b/students/Zach Cooper/lesson06/Description of Changes Made to good_perf.py
@@ -1,0 +1,46 @@
+1. Initial poor_perf.py output data and time it took to run the loop one time
+	C:\Users\Zach\UWPYTHON\PythonCert220Assign\students\Zach Cooper\lesson06>python -m cProfile poor_perf.py
+	{'2013': 1025, '2014': 1013, '2015': 1111, '2016': 1023, '2017': 2029, '2018': 0}
+	'ao' was found 63395 times
+	20.464950117999997
+         	1085304 function calls (1085280 primitive calls) in 20.474 seconds
+2. Changed a typo for the year in year_count. Intitially returned '2018': 0 but now returns '2018': 5811.
+3. Added a simple test to assert that my year_count and 'ao' count where equal.
+4. Added code to skip the header. The header was breaking the code when we are trying to convert the date string
+   using datetime.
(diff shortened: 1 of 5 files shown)


students/Zach Cooper/lesson06/Description of Changes Made to good_perf.py
```diff
@@ -0,0 +1,46 @@
...
+	 1000001   17.639    0.000   32.750    0.000 _strptime.py:318(_strptime)
+               1    0.000    0.000    0.001    0.001 _strptime.py:49(__init__)
+         1000001    2.933    0.000   35.683    0.000 _strptime.py:574(_strptime_datetime)
+   
+   There is two ways to think about this. If you realistically need to only iterate and use this file once or twice
+   then using datetime is a good idea because we are not always guaranteed a clean set of dates that are all in
+   the same date format as this file is so using datetime output a clean set of data. On the other hand, its really
+   slow and if you know that all of your dates are intially in the correct format, it would be quickest to
+   just to leave them in string format as the output time will be substantially quicker. It is all up the the client
+   and what they are looking for.
```
codedragon: I pretty much agree with your assessment. I think it has more to do with the date integrity than how many times you will use it, however. If it is very important to get the dates exactly right and they may not be, then you will spend the extra time every time, but if you are reasonably sure they are correct, then you probably won't ever check them. It also depends on how much data you have, how long is acceptable, and what they are doing with the data, etc. So, yes, comes down to what the client is looking for.

students/Zach Cooper/lesson06/alt_good_perf.py
```diff
@@ -0,0 +1,57 @@
...
+        year_count = {}
+
+        # Return 'ao' count start at 0
+        found = 0
+
+        # Loop through date strings between 2012 and 2018 and return count for each year
+        # Date column is index 5 in file and [-4:] parseson just year of the date string
+        for row in reader:
+            if "2012" < row[5][-4:] < "2019":
+                year_count[row[5][-4:]] = year_count.get(row[5][-4:], 0) + 1
```
codedragon: This is the same as what you have, a little shorter notation:
```
year_count[row[5][-4:]] += 1
``

PR #85, opened 2019-02-26
Added assignment files under lesson07

dtingg:
Per the assignment requirements, I changed linear.py so that it returns two tuples, one for customers and one for products.

My version using multithreading is in parallel.py.  It seems a little bit faster, but not by much.  I had a difficult time returning values from the threads, so I passed in a dictionary.  I also wasn't sure if I had to add any locking, so I assume the GIL is taking care of that for me.  :-)  

I made some tests for both files and all of them pass.  My notes are in findings.txt.  Thank you!

### DIFF
--- a/students/DiannaTingg/lessons/lesson07/assignment/findings.txt
+++ b/students/DiannaTingg/lessons/lesson07/assignment/findings.txt
@@ -9,13 +9,13 @@ Each tuple contains: number of records processed, record count in the database p
 I'm using larger csv files for products, customers, and rentals.  Each file has 9,999 records plus a header row.
 I automatically clear the MongoDB each time the program exits.
 
-Using timeit, my linear.py file takes 3.458133672 seconds to run once.  It takes 29.399097648999998 seconds to run 10 times.
+Using timeit, my linear.py file takes 2.573601484 seconds to run once.  It takes 25.389773754 seconds to run 10 times.
 I wrote some tests in test_linear.py and they all pass.
 
 I'm going to use multithreading in my parallel.py file to see if I can get a performance improvement.
 I will read all three csv files at the same time (in parallel).
 
-Using timeit, my parallel.py file takes 2.166639083 seconds to run once.  It takes 19.358340443 seconds to run 10 times.
+Using timeit, my parallel.py file takes 1.679744485 seconds to run once.  It takes 18.042013637 seconds to run 10 times.
 I wrote some tests in test_parallel.py and they all pass.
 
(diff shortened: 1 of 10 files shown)


students/DiannaTingg/lessons/lesson07/assignment/findings.txt
```diff
@@ -0,0 +1,22 @@
...
+I wrote some tests in test_linear.py and they all pass.
+
+I'm going to use multithreading in my parallel.py file to see if I can get a performance improvement.
+I will read all three csv files at the same time (concurrently).
+
+Using timeit, my parallel.py file takes 1.679744485 seconds to run once.  It takes 18.042013637 seconds to run 10 times.
+I wrote some tests in test_parallel.py and they all pass.
+
+I keep getting different times from timeit, but in general, parallel.py seems to be faster than linear.py, so I would recommend using that program instead.
+The program never seems to fail due to contention.
```
codedragon: Whether or not there is much speed increase depends on how mongodb is implemented. Since all parallel calls are going to mongodb, if it then takes them more or less linearly, it won't make  a huge difference. The fact that it improved the time a bit, means it is a bit better that. Probably if we were using a network database with lots of workers, we would see more improvement.

students/DiannaTingg/lessons/lesson07/assignment/parallel.py
```diff
@@ -0,0 +1,192 @@
...
+               threading.Thread(target=_add_bulk_data, args=(results_dict, customers, directory_name, customers_file)),
+               threading.Thread(target=_add_bulk_data, args=(results_dict, rentals, directory_name, rentals_file))]
+
+    for t in threads:
+        t.start()
+
+    for t in threads:
+        t.join()
+
+    return [results_dict["customers"], results_dict["products"]]
```
codedragon: Using a dictionary is a fine solution. If you are using lots of threads, or threads need to share data, queue or dequeue is the answer.

students/DiannaTingg/lessons/lesson07/assignment/test_parallel.py
```diff
@@ -0,0 +1,94 @@
+"""
```
codedragon: Could have probably used the same test file, and imported either parallel or linear, depending on which you are testing. If this was a real project, with both a linear and parallel option, you would want to figure out how to automate running the same tests on both files, so the same functionality is always preserved. Probably the easiest way to do this is subclass one from the other, so you only write the tests once, but it is clear that both run.

PR #86, opened 2019-02-26
Lesson 7

vanulor:
I found with at least the way I did threading that linear was faster.  I timed it running 100 times each and it was faster to not run it using parallel.  I implemented a possible collision, but I think the method I am using ensures thread safety but this would be what would cause a issue if show_available product finished before the data import then the data would collide and there would be a issue.  Based on my finding I wouldn't recommend using multi-threading for this project.


students/BrianNisonger/Lesson7/assignment/parallel.py
```diff
@@ -0,0 +1,55 @@
+from multiprocessing.pool import ThreadPool
+from database import import_data
+from database import show_available_products
+
+
+def run_parallel():
+    """
+    Run the import data in parallel
+    """
+    pool = ThreadPool()
```
codedragon: Aha! I think the problem is you have to initiate it with how many threads you want. The default is probably one. Try:
```
pool = ThreadPool(3)
```
and see if that speeds things up.

students/BrianNisonger/Lesson8/assignment/inventory.py
```diff
@@ -0,0 +1,34 @@
...
+        try:
+            customer_writer.writerow(
+                [args[1], args[2], args[3], f'{args[4]:.2f}'])
+        except IndexError:
+            logging.error("Missing values")
+            raise (IndexError)
+        except ValueError:
+            logging.error("Not a number")
+            raise (ValueError)
+    return None
```
codedragon: Not quite what they had in mind, but such piss poor instructions, how could you know? The idea was that you actually had all five things as inputs (rather than *args), and then in the single_customer you could use the add_furniture to loop through and add each line from the one csv file to the other. You would use partial so that you wouldn't have to add the customer name and the file each time. It doesn't really save you much, since it would be one line, but it was to demonstrate how to use partial to change the signature of a function. You got the closure right, though, and that is much more important than partial, which you may or may not ever use again. :)

PR #87, opened 2019-02-26
Assignment 07

CodeChameleon:
Assignment 7 files. Earlier in the course, the Sqlite3 db was referred to as the "Customer" db and the MongoDB was the "Products" db. I wasn't sure if I needed to multithread both of those to run together or just MongoDB. I did both...for the experience?

I did get some improvement out of using threads on the Sqlite3 inserts (batches of 120 records up to 10,000 records total). The MongoDB inserts didn't really see much improvement though.

I used threadpools and thought it was interesting that performance (time to complete) got worse if I used too many threads on the Sqlite3 side. 3 threads worked best. Probably an issue with having the DB running on the same machine requesting the inserts.


students/DrewSmith/lessons/lesson07/assignment/parallel.py
```diff
@@ -0,0 +1,277 @@
...
+    start = datetime.datetime.now()
+    parent_path = Path(__file__).parent
+
+    products_starting = get_prodcuts_total_record_count()
+    customers_starting = get_customers_record_count()
+
+    products_thread = pool.apply_async(import_product_data, (parent_path, "products.csv", "contacts.csv", "rentals.csv"))
+    customers_thread = pool.apply_async(bulk_add_customers, (convert_csv("customers.csv"), ))
+    pool.close()
+    pool.join()
```
codedragon: Interesting, so you are threading for mongo and sqlite, and then threading again for each of those threads? I'm not sure how this would affect things. I'd be curious how the sqlite and mongo do individually. Or for that matter, threading only the top level. Did you try any of those? I'm not sure how threading a thread goes, in general, but I'm definitely intrigued.

students/DrewSmith/lessons/lesson07/assignment/linear.py
```diff
@@ -0,0 +1,215 @@
...
+            customer['credit_limit'] = float(customer['credit_limit'])
+        except ValueError as err:
+            logging.error(err)
+            raise ValueError(f"Invalid value: record {index}: " \
+                f"invalid credit_limit: '{customer['credit_limit']}'")
+
+        if not isinstance(customer['status'], bool):
+            text = f"Invalid value: record {index}: status is not a bool: '{customer['status']}'"
+            logging.error(text)
+            raise ValueError(text)
```
codedragon: I would consider putting the error checking in a separate function to make this function smaller and more readable.

students/DrewSmith/lessons/lesson07/assignment/linear.py
```diff
@@ -0,0 +1,215 @@
...
+            counts[1] = len(records)
+            database["customers"].insert_many(records)
+
+        with open(directory / rentals_file, 'r') as file:
+            csv_file = csv.DictReader(file, fieldnames=("product_id", "user_id"))
+            records = [record for index, record in enumerate(csv_file) if index > 0]
+            counts[2] = len(records)
+            database["rentals"].insert_many(records)
+
+    return (tuple(counts), tuple(errors))
```
codedragon: import_product_data is very long. I would consider breaking it up into a 2 or 3 functions

PR #89, opened 2019-03-02
ericstreit pull request

ericstreit:
I think it is has been a while that I made a pull request. A few assignments here.


students/ericstreit/lessons/lesson02/assignment/code/charges_calc.py
```diff
@@ -5,43 +5,135 @@
...
     with open(filename) as file:
+        logging.debug(f'Opening {filename}')
         try:
             data = json.load(file)
-        except:
+        except Exception as except_error:
+            print(except_error)
+            logging.error(except_error)
             exit(0)
     return data
```
codedragon: Nice improvements. :)

students/ericstreit/lessons/lesson02/assignment/code/charges_calc.py
```diff
@@ -5,43 +5,135 @@
...
+                            f"Price: {except_error}")
+            # pass
+        try:
             value['unit_cost'] = value['total_price'] / value['units_rented']
-        except:
-            exit(0)
+            logging.debug(f"Updating {value} unit cost to: {value.get('unit_cost')}")
+        except Exception as except_error:
+            logging.error(f': {key} : Error calculating Unit Cost: {except_error}')
+            # pass
```
codedragon: It is a lot of try/excepts, but that is what happens when cleaning data. Would be nice to catch the precise exception (like KeyError or whatever) rather than the generic Exception. This is because Exception catches everything, and sometime this is an unexpected error that you would prefer to throw, if you knew about it.

students/ericstreit/lessons/lesson02/assignment/code/charges_calc.py
```diff
@@ -5,43 +5,135 @@
...
+        logger.setLevel(logging.DEBUG)
+        logging.info(f'Configuring debug parameters for level {level} DEBUG')
+
+    elif level == 2:
+        logger.setLevel(logging.WARNING)
+        logging.info(f'Configuring debug parameters for level {level} WARNING')
+
+    elif level == 3:
+        logger.setLevel(logging.ERROR)
+        logging.info(f'Configuring debug parameters for level {level} ERROR')
```
codedragon: A little irony here?

students/ericstreit/lessons/lesson03/assignment/basic_operations.py
```diff
@@ -0,0 +1,121 @@
+"""
+#Lesson03
+#Customer Database Exercise ##
+"""
+
+#!/usr/bin/env python3
+#import stuff here
+from create_db import *
+from db_model import *
```
codedragon: The peewee model set a poor example. Please avoid import *. It makes the name space very messy, and it is hard to figure out where things came from. For example, if something strange happened when running Customer.create, how would you know what file to find it in? What if you hadn't looked at this code in 6 months would you still remember? Yes, of course you can search for it, but better to be explicit.

students/ericstreit/lessons/lesson03/assignment/create_customers.py
```diff
@@ -0,0 +1,54 @@
...
+                    status = account__status,
+                    credit_limit = credit__limit)
+            customer.save()
+            logging.info('Database add successful')
+
+    except Exception as e:
+        logging.info(f'Hm, we had an error here?')
+        logging.info(e)
+    finally:
+        db.close()
```
codedragon: It would make more sense to put the try block just around the piece you think might cause an error, so Customer.create and customer.save, probably, rather than including everything. And then you let the user know what happened. I can see you already realize you may have errors you have not anticipated. ;-) So, if there is a problem with connecting to the database, rather than a problem with saving a customer, you probably want you customer to know that the database can't be reached at the moment, rather than just swallowing that error. If the problem is creating a customer, than you want them to know that there was something wrong with the information for that customer, but atm, in both cases, you are swallowing the error, and the user would not have any idea anything even went wrong.

students/ericstreit/lessons/lesson03/assignment/db_model.py
```diff
@@ -0,0 +1,82 @@
...
+# class PersonNumKey(BaseModel):
+#     """
+#         This class defines Person, which maintains details of someone
+#         for whom we want to research career to date.
+#     """
+#     logger.info('An alternate Person class')
+#     logger.info("Note: no primary key so we're give one 'for free'")
+#     person_name = CharField(max_length = 30)
+#     lives_in_town = CharField(max_length = 40)
+#     nickname = CharField(max_length = 20, null = True)
```
codedragon: I think assignment three was suppose to have tests. Please write tests!

students/ericstreit/lessons/lesson06/assignment/good_perf.py
```diff
@@ -0,0 +1,62 @@
...
+        found = 0
+        for row in reader:
+            lrow = list(row)
+            if "ao" in lrow[6]:
+                found += 1
+                if lrow[5] > '00/00/2012':
+                    new_ones.append((lrow[5], lrow[0]))
+            else:
+                if lrow[5] > '00/00/2012':
+                    new_ones.append((lrow[5], lrow[0]))
```
codedragon: I would probably have "ao" and the dates be separate if statements, then you don't have duplicate code. Also, I wonder if it takes more time to put new_ones in a list and then put it in the dictionary. Why not just add them directly into the dictionary?

students/ericstreit/lessons/lesson08/assignment/inventory.py
```diff
@@ -0,0 +1,98 @@
...
+    add_furniture('invoice01.csv', 'King Kong', 'FB04', 'Fruit Basket', 100.00)
+    add_furniture('invoice01.csv', 'Godzilla', 'AA48', 'Antacid Basket', 25.00)
+    add_furniture('invoice01.csv', 'Mothra', 'CB10', 'Clothes Basket', 33.00)
+    add_furniture('invoice01.csv', 'Mechazilla', 'SP55', 'Spare Parts Basket', 147.00)
+    add_furniture('invoice01.csv', 'Monster X', 'HB94', 'Hat Basket', 16.00)
+
+    #assertions
+    king_kong = partial(single_customer_partial, new_invoice_file =
+                        'king_kong_invoice.csv')
+    assert king_kong('King Kong') == 'Invoice king_kong_invoice.csv created for King Kong.'
```
codedragon: Oh, I see. The directions were terrible. Partial should be in the function. Note your assert checks to see that you returned something, but doesn't check to see if there is anything in the csv file. So, we don't really know if the function did what it was suppose to do.

students/ericstreit/lessons/lesson08/assignment/inventory.py
```diff
@@ -0,0 +1,98 @@
...
+                if customer_name in row:
+                    # convert the iteration into a list object in order to send
+                    # back into a csv
+                    data = [row]
+                    with open(new_invoice_file, 'a') as csvfile2:
+                        writer = csv.writer(csvfile2, delimiter=',', quotechar='"')
+                        writer.writerows(data)
+        return(f'Invoice {new_invoice_file} created for {customer_name}.')
+        
+    return single_customer_lookup
```
codedragon: So, the idea was to have a closure like you have, but the outer function takes the customer's name and the new invoice file name, and the inner function takes the old invoice file name and uses partial to call add_furniture with just the rental information, since the customer name and invoice file aren't changing, and so it adds the rental information from the old invoice file name and adds it to the new invoice. Clear as mud? Anyway, nice closure. :)

students/ericstreit/lessons/lesson08/assignment/inventory.py
```diff
@@ -0,0 +1,98 @@
...
+            for row in reader:
+                if customer_name in row:
+                    # convert the iteration into a list object in order to send
+                    # back into a csv
+                    data = [row]
+                    with open(new_invoice_file, 'a') as csvfile2:
+                        writer = csv.writer(csvfile2, delimiter=',', quotechar='"')
+                        writer.writerows(data)
+        # I think I need to return something here?
+        return(f'Invoice {new_invoice_file} created for {customer_name}.')
```
codedragon: I don't understand. Where is partial? It should still be a closure, btw, partial is just the name of the function you are using to make the closure more interesting. Or something. In this case, it doesn't really do much as far as saving code writing, but it was suppose to show you how it works, and hopefully you would find better use cases for it. At any rate, closures are much more important, and it looks like you made one successfully. :) Well, above, not here, where confusion reigns. By no fault of yours.

PR #91, opened 2019-03-03
submit lesson05 hw

ShujinCao:


students/ShujinCao/lesson05/assignment/database.py
```diff
@@ -0,0 +1,78 @@
...
+def upload_data(x, data):
+    mongo = MongoDBConnection()
+    result = []
+
+    with mongo:
+        for i in range(len(data)):
+            data = data.astype(str)
+            value = dict(data.loc[i])
+            result.append(value)
+        x.insert_many(result)
```
codedragon: This is a little confusing. Why are you re-assigning data as you go through the loop? For that matter, why are you doing it every time you go through the loop. Data isn't changing, so you should be able to do it once, outside of the loop, and then iterate. When you do things that are difficult to figure out, it is good to add comments to explain why you are doing what you are doing. This way other people (or you in six months) will be able to figure out what you did and why.

students/ShujinCao/lesson05/assignment/test_database.py
```diff
@@ -0,0 +1,16 @@
...
+                  'prd010']
+
+def test_list_renter_info():
+    results = db.list_renter_info("prd005")
+    assert [*results] == ['user001', 'user003']
+
+
+def test_list_available_products():
+    results = db.list_available_products()
+    assert [k for k in results.values()][4]['description'] == 'Portable heater'
```
codedragon: Nice start to testing. It is good to have tests not dependent on each other. There is no guarantee that tests are going to be run in a particular order, and it makes it so you can run a single test, which can be very helpful. Fixture can help with this. 
http://doc.pytest.org/en/latest/fixture.html

students/ShujinCao/lesson05/assignment/database.py
```diff
@@ -0,0 +1,78 @@
...
+
+    mongo = MongoDBConnection()
+
+    with mongo:
+        db = mongo.connection.media
+        product_df = pd.read_csv(f_product)
+        customer_df = pd.read_csv(f_customer)
+        rental_df = pd.read_csv(f_rental)
+
+        x_product, x_customer, x_rental = db["products"], db["customers"], db["rentals"]
```
codedragon: Great use of multiple assignments.

students/ShujinCao/lesson06/assignment/good_perf.py
```diff
@@ -0,0 +1,53 @@
...
+
+    print(year_count)
+
+    found = df[" sentence"].str.contains("ao").sum()
+
+    print(f"'ao' was found {found} times")
+    end = datetime.datetime.now()
+    print(timeit.default_timer() - start_time)
+
+    return (start, end, year_count, found)
```
codedragon: I'll be interested in how using pandas compares with other methods. It seems like you are iterating once, instead of the two for loops of the original, so it should definitely improve things.

PR #92, opened 2019-03-04
Lesson 07 and 08 Pull Request

zconn:
Finished linear and parallel reporting and created currying function for assignment 8


codedragon:
Hi Zach,
I'm wondering if you missed some commits in this pull request. I think some things are missing from your threads, and I don't see the currying at all.

PR #93, opened 2019-03-04
Lesson6 completed

Misseswaddles:


students/cfarris/lessons/lesson06/assignment/good_perf.py
```diff
@@ -0,0 +1,85 @@
...
+            if new[0][6:] == '2014':
+                year_count["2014"] += 1
+            if new[0][6:] == '2015':
+                year_count["2015"] += 1
+            if new[0][6:] == '2016':
+                year_count["2016"] += 1
+            if new[0][6:] == '2017':
+                year_count["2017"] += 1
+            if new[0][6:] == '2018':
+                year_count["2017"] += 1
```
codedragon: Well done. You found the big one, opening the file twice. Next would be to not append to a list and then add to the dictionary, but instead put directly into the dictionary.

Misseswaddles:
Maria,

I only wanted you to look at Lesson 6 for now. I am finishing up Lesson7
and hope to submit a little later today. I am going to save any edits and
revisions on past assignments for the break--no need to look at those.

Thanks!
Carol

PR #96, opened 2019-03-06
submitting files for homework 8

kmonster462:
Submitting homework 8 and all relevant files associated with it.

### DIFF
--- a/students/Kate_Koehn/Lesson08/inventory.py
+++ b/students/Kate_Koehn/Lesson08/inventory.py
@@ -1,0 +1,54 @@
+'''
+To summarize: 1. You will create a program to initially create, and subsequently update, a CSV file that lists
+which furniture is rented to which customer (to replace use of the spreadsheet mentioned above).
+2. You will create additionally functionality that will load individual customers rentals.
+'''
+
+import csv
+from functools import partial
+
+
+def add_furniture(invoice_file="", customer_name="", item_code="", item_description="", item_monthly_price=0.00):
+    with open(invoice_file, 'a+') as csv_file:
+        writer = csv.writer(csv_file)
+        row = customer_name, item_code, item_description, item_monthly_price
+        writer.writerow(row)
+
+
+def test_add_data():
+    """
+    function to add test data using add_furniture function
+    """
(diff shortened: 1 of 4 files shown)


students/Kate_Koehn/Lesson08/inventory.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def add_furniture(invoice_file="", customer_name="", item_code="", item_description="", item_monthly_price=0.00):
+    with open(invoice_file, 'a+') as csv_file:
+        writer = csv.writer(csv_file)
+        row = customer_name, item_code, item_description, item_monthly_price
+        writer.writerow(row)
+
+
+def test_add_data():
```
codedragon: Be careful naming something that is not actually a test with "test_" If you have run a test runner, it might just find it and run it, which is probably not what you want.

students/Kate_Koehn/Lesson08/inventory.py
```diff
@@ -0,0 +1,54 @@
...
+    def add_rentals(rental_file):
+        with open(customer_file) as rental_csv:
+            reader = csv.reader(rental_csv)
+
+            add_item = partial(add_furniture, customer_name=customer_name, invoice_file=rental_file)
+
+            for row in reader:
+                add_item(item_code=row[1], item_description=row[2], item_monthly_price=row[3])
+
+    return add_rentals
```
codedragon: Closure looks great. Remember in the doc string to be explicit that it will return a function. Functions like this should be very explicit, outer function takes these variables, returns a function that takes these variables.

students/Kate_Koehn/Lesson08/inventory.py
```diff
@@ -0,0 +1,54 @@
...
+    add_furniture("susans_stuff.csv", "Susan Wong", "GH9S", "Lamp", 200.00)
+    add_furniture("susans_stuff.csv", "Susan Wong", "L4J9", "Table", 4000.00)
+
+
+if __name__ == "__main__":
+    test_add_data()
+    test_add_susan()
+
+    susans_crap = single_customer("Susan Wong", "susans_stuff.csv")
+    susans_crap("invoice_file.csv")
```
codedragon: haha. susans_crap? what? ;-)

PR #99, opened 2019-03-07
Lesson08 assignment

ktbzzz:


students/JaeKim/lesson08/assignment/inventory.py
```diff
@@ -0,0 +1,25 @@
...
+                new_csv.write("{},{},{},{}".format(customer_name, item[1], item[2], item[3]))
+    return another_function
+
+if __name__ == "__main__":
+    add_furniture("invoice_file.csv", "Elisa Miles", "LR04", "Leather Sofa", "25.00")
+    add_furniture("invoice_file.csv", "Edward Data", "KT78", "Kitchen Table", "10.00")
+    add_furniture("invoice_file.csv", "Alex Gonzales", "BR02", "Queen Mattress", "17.00")
+
+    create_invoice = single_customer("Susan Wong", "invoice_file.csv")
+    create_invoice("test_items.csv")
```
charlesrobison: Note this is creating test_items.csv with Susan Wong as the customer for each item of inventory.  Don't think this is correct.  Program is running well though.

ktbzzz: This assignment was a bit confusing, and I added every item as "Susan Wong" based off my interpretation which seems to be incorrect.

What I "interpreted" from Marias message on Slack:
create_invoice = single_customer("Susan Wong", "rented_items.csv")
create_invoice("test_items.csv")

So, using create_invoice() will, in this case, add all items in test_items.csv to rented_items.csv under the name "Susan Wong".

codedragon: This exercise was definitely confusing, many apologies for that. I wasn't sure what he was asking when we began, either. Honestly, from my perspective, I don't care which way you move the lines. The important part is that you have a closure, that returns a function, that move stuff from one file to another. That part you have, and it looks good. The part that is missing is using partial, which is less important, but would be good to play with. Partial takes a function and "fixes" one or more of the input variables, so you don't have to re-enter the same thing over and over. The way you are using add_furniture in your code above won't work, because add_furniture has 5 input variables and you are giving it 4. If you use partial to change the signature of add_furniture in your closure, then you will be able to give your new function less than 5, specifically the directions said to change the signature so that the name and file are fixed, since these will always be the same when using your closure, and the actual furniture information are the only things entered every time through the loop. Hope this helps a little. Great job making a closure!

PR #100, opened 2019-03-07
Lesson07 & Lesson08 completed

Misseswaddles:
Pull for Lesson07. All files included.


Misseswaddles:
Note,
For lesson 08, pylint complained about my partial:
Argument 'invoice_file' passed by position and keyword in function call (redundant-keyword-arg)
However, it seemed like this was a reported problem with using partials and pylint. I may have misunderstood, however. Let me know if I am passing parameters around incorrectly with my partial. Thanks in advance!

students/cfarris/lessons/lesson07/assignment/linear.py
```diff
@@ -0,0 +1,226 @@
...
+        count_before_running = collection.count()
+        for row in csv_reader:
+            if header:
+                header = False
+                first_line = [value.strip() for value in row]
+            else:
+                num_processed += 1
+                new_dict = {}
+                str_row = []
+                str_row = [value.strip() for value in row]
```
codedragon: You don't have to initialize variables with an empty list, if you are doing a comprehension. The comprehension is all you need!

students/cfarris/lessons/lesson07/assignment/test_parallel.py
```diff
@@ -0,0 +1,156 @@
...
+        self.host = host
+        self.port = port
+        self.connection = None
+
+    def __enter__(self):
+        self.connection = MongoClient(self.host, self.port)
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.connection.close()
```
codedragon: You know you don't have to copy and paste MongoDBConnection everywhere, just import it. :) In fact in this exercise, you have a lot of copied code. Get in the habit of using stuff you already have, instead of copying it. How much of this test file is different from the linear one? Could you just run it twice with different files?

students/cfarris/lessons/lesson07/assignment/parallel.py
```diff
@@ -0,0 +1,243 @@
...
+    """
+    For each row to be entered into a collection,
+    this checks to see if it is already there.
+    I cannot get this to work.
+    """
+
+    if collection.find({unique_identifier_key:
+                        unique_identifier_value}).count() > 0:
+        return False
+    return True
```
codedragon: The unique identifier is always unique, so you won't find duplicates. The way to check for redundancy with mongo is to assign the _id to your product or customer id as you enter it in the database. Then, mongo will complain if you try to put in the same one twice. If you let mongo assign the _id, mongo will assign a unique id to the second entry of the same thing, and not pay any attention that everything else is the same.

students/cfarris/lessons/lesson08/assignment/inventory.py
```diff
@@ -0,0 +1,58 @@
...
+                                  invoice_file=invoice_file,
+                                  customer_name=customer_name,)
+
+            for row in csv_reader:
+                store_increment = store_increment + 1
+                add_rentals(row[0],
+                            row[1],
+                            row[2])
+
+    return add_cust_rentals_to_invoice
```
codedragon: store_increment is not crucial to your function, and you're not returning it, so I wouldn't include it.
I see why pylint is complaining. You have too many things called invoice file. And, it is true, that in this case they are actually all the same file, but it doesn't have to be that way. You could call the input to single_customer general_invoice or something, then in the partial it would be invoice_file=general_invoice. But, I agree that this is one of those fuzzys that you would decide with the customer so that it is clear for them, since that is most important. If you decide it makes sense for the input to both functions to have the same name, tell pylint to ignore that line.

PR #101, opened 2019-03-08
Assignment 9

vanulor:
I update the context manager in database.py to test for configuration errors and added a test to test_database.py to make sure it was catching it.  I also only close the connection if it was successfully created.  I added a decorator to calc_charges to enable logging to be turned on and off.  I created jpgdiscover.py It uses a dictionary instead of a list which was cleaner for me.  It does require a dict to be created before the function is run.


students/BrianNisonger/Lesson9/jpgdiscover.py
```diff
@@ -0,0 +1,30 @@
...
+        file_path = os.path.join(directory, file)
+        if os.path.isdir(file_path):
+            jpegdiscovery(file_path)
+        elif file.endswith(".png"):
+            if directory in file_dict.keys():
+                file_dict[directory].append(file)
+            else:
+                file_dict[directory] = [file]
+
+    return [[file, file_dict[file]] for file in file_dict]
```
codedragon: Not sure that it is really cleaner, if the user has to instantiate an empty dictionary outside of your function. I think even with the dictionary, if you had file_dict as a keyword argument, assigned to None, like we did in class, you wouldn't need to have it in the outside scope.

PR #103, opened 2019-03-10
Adding parts 1 & 3 of HW 9

kmonster462:
Part 1 still stops all logging vs just stopping the logging of the decorated function. Still working on that.

### DIFF
--- a/students/Kate_Koehn/Lesson09/jpgdiscover.py
+++ b/students/Kate_Koehn/Lesson09/jpgdiscover.py
@@ -1,0 +1,29 @@
+"""
+Recursively go through directories looking for pngs
+"""
+
+import os
+
+
+def jpg_discovery(directory, png_paths=None):
+    """
+    Search directory recursively for all png files and add to dict
+    """
+    files = os.listdir(directory)
+    for file in files:
+        file_path = os.path.join(directory, file)
+        if os.path.isdir(file_path):
+            jpegdiscovery(file_path)
+        elif file.endswith(".png"):
+            if directory in file_dict.keys():
+                file_dict[directory].append(file)
+            else:
+                file_dict[directory] = [file]
+
+
+if __name__ == "__main__":
+    #Format output
+    file_dict = {}
+    jpegdiscovery(os.getcwd())
+    file_list = [[file, file_dict[file]] for file in file_dict]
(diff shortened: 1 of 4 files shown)


students/Kate_Koehn/Lesson09/jpgdiscover.py
```diff
@@ -0,0 +1,29 @@
...
+    for file in files:
+        file_path = os.path.join(directory, file)
+        if os.path.isdir(file_path):
+            jpegdiscovery(file_path)
+        elif file.endswith(".png"):
+            if directory in file_dict.keys():
+                file_dict[directory].append(file)
+            else:
+                file_dict[directory] = [file]
+
```
codedragon: What's up with the dictionary? If you want to use a dictionary, you should be able to add it as another keyword argument, initialize when it is None, and not have to declare it outside of the function, which is not very user friendly.

students/Kate_Koehn/Lesson09/logging_decorators.py
```diff
@@ -0,0 +1,147 @@
...
+    """
+
+    def decorated_func(*args, **kwargs):
+
+        logging.disable(logging.CRITICAL)
+        returned_value = func(*args, **kwargs)
+        logging.disable(logging.NOTSET)
+        return returned_value
+
+    return decorated_func
```
codedragon: But, heh, you got the decorator working. :)

students/Kate_Koehn/Lesson10/database.py
```diff
@@ -0,0 +1,166 @@
...
+        print('{:s} function took {:.3f} ms'.format(func.__name__, (time2-time1)*1000.0))
+        with open("timing_output.txt", "a") as output_file:
+            output_file.write('{:s} function took {:.3f} ms\n'.format(func.__name__, (time2-time1)*1000.0))
+        return end
+    return wrap
+
+
+@timing
+class MongoDBConnection():
+    """MongoDB Connection"""
```
codedragon: Using the timing decorator on this class is kind of weird. You need special decorators for classes, otherwise you are just timing how long it took to instantiate the class, which is not terribly useful. What you would really want to know is how long it took to connect to the database .

PR #104, opened 2019-03-11
Assignment 09 Updates

CodeChameleon:
1. I added a command argument "-l" so I could specify which method(s) I wanted to disable logging for. I set the decorator for the appropriate method(s) in the __name__ == '__main__' section. This seemed to work, disabling whichever method I entered (options are "load", "calc", or "save".

2. I wasn't sure which "features" I should implement in the context manager, but I added some error handling in the __enter__ and checked that the connection exists before closing it.

3. I like using pathlib, so I wanted to attempt to switch what we did in class to pathlib. Pretty much the same method. I used the iterdir() method which includes the entire path instead of just the file/directory name.


students/DrewSmith/lessons/lesson09/assignment/charges_calc.py
```diff
@@ -0,0 +1,185 @@
...
+    ARGS = parse_cmd_arguments()
+
+    # Selective method logging deactivate
+    if ARGS.log_off is not None:
+        if 'load' in ARGS.log_off.lower():
+            load_rentals_file = disable_method_logging(load_rentals_file)
+        if 'calc' in ARGS.log_off.lower():
+            calculate_additional_fields = disable_method_logging(calculate_additional_fields)
+        if 'save' in ARGS.log_off.lower():
+            save_to_json = disable_method_logging(save_to_json)
```
codedragon: Interesting approach. It works well since this file is all about using the command line, and this is doing more than the assignment asked for (I think they were assuming that the switch in the args would be more like an on/off, and would only affect the decorated functions, but would not decide which ones were decorated). But, this is an appropriate solution, given the criteria. Note that if you actually wanted to be able to import these functions in to another file, it would be better to use the syntactic sugar that is the @ symbol, but then you would have a whole new set of problems regarding who is controlling those decorators. :)

students/DrewSmith/lessons/lesson09/assignment/database.py
```diff
@@ -0,0 +1,150 @@
...
+            except pymongo.errors.ConnectionFailure as error:
+                logging.error(f"MongoDB ConnectionFailure: {error}")
+                raise            
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        ''' Close connection '''
+        if self.connection is not None:
+            self.connection.close()
+        return False
```
codedragon: Cool. In the end, I'm not sure this is worth it, as in I'm not sure it is really adding much value. You could accomplish pretty much the same thing, by adding a logging statement in __exit__ that logged anything if there was an exception. I guess it would be broader in the exceptions it caught, but that would probably not be a bad thing.
But it was a vague assignment, and you got some good practice working with context managers. :)

students/DrewSmith/lessons/lesson09/assignment/jpgdiscover.py
```diff
@@ -0,0 +1,29 @@
...
+    current_list = [directory, []]
+    for path in directory.iterdir():
+        if path.is_dir():
+            jpgdiscovery(path, png_paths)
+        elif path.name.endswith(".png"):
+            current_list[1].append(path.name)
+
+    if current_list[1]:
+        png_paths.extend(current_list)
+    return png_paths
```
codedragon: Very nice. I knew it would be straightforward with Paths, but I haven't played around with them enough to be able to do it off the cuff in class. Glad someone figured it out! :)

students/DrewSmith/lessons/lesson09/assignment/charges_calc.py
```diff
@@ -0,0 +1,185 @@
...
+        print("Output file complete")
+
+def parse_cmd_arguments():
+    ''' Processes commandline arguments from the user '''
+    parser = argparse.ArgumentParser(description='Process some integers.')
+    parser.add_argument('-i', '--input', help='input JSON file', required=True)
+    parser.add_argument('-o', '--output', help='ouput JSON file', required=True)
+    parser.add_argument('-d', '--debug', help='ouput JSON file', required=False)
+
+    parser.add_argument('-l', '--log_off', help='ouput JSON file', required=False)
```
codedragon: Why are the help for all of your outputs the same? You would definitely want to include help for log_off, since it is unusual.

PR #105, opened 2019-03-11
Lesson05 completed (mongodb)

ericstreit:
A bunch of Lesson10 as well but still need to provide test results on the timing

### DIFF
--- a/students/ericstreit/lessons/lesson04/assignment/create_customers.py
+++ b/students/ericstreit/lessons/lesson04/assignment/create_customers.py
@@ -1,0 +1,54 @@
+"""
+#Lesson03
+#Customer Database Exercise ##
+"""
+
+# !/usr/bin/env python3
+# import stuff here
+from db_model import *
+import logging
+logging.basicConfig(level=logging.DEBUG)
+# global variables here
+
+# define function
+def create_customers(first__name, last__name, home__address, phone__number, email__address,
+                     account__status, credit__limit):
+    """
+    Description of function
+
+    :param arg1: The first very important parameter. And a bit about
+                 what it means.
+    :param arg2: The second very important parameter. And now some
+                 description of how this is used
+    etc
+    """
+    try:
+        with db.transaction():
+            logging.info(f'opening database: {db}')
+            customer = Customer.create(
+                    first_name = first__name,
+                    last_name = last__name,
+                    home_address = home__address,
+                    phone_number = phone__number,
+                    email_address = email__address,
+                    status = account__status,
+                    credit_limit = credit__limit)
+            customer.save()
+            logging.info('Database add successful')
+    except Exception as e:
+        logging.info(f'Hm, we had an error here?')
+        logging.info(e)
+    finally:
+        db.close()
+
+
+def customer_query(k):
+    db.connect()
+    r = Customer.select().where(Customer.first_name == k).get()
+    print(r)
+    db.close()
+
+# for testing
+if __name__=="__main__":
+    pass
+    #customer_query(1)
--- a/students/ericstreit/lessons/lesson04/assignment/db_model.py
+++ b/students/ericstreit/lessons/lesson04/assignment/db_model.py
@@ -1,0 +1,56 @@
+"""
+    Simple database example with Peewee ORM, sqlite and Python
+    Here we define the schema
+    Use logging for messages so they can be turned off
+
+"""
+import logging
+from peewee import *
+
+logging.basicConfig(level=logging.INFO)
+# logger = logging.getLogger(__name__)
+
+logging.info('Here we define our data (the schema)')
+logging.info('First name and connect to a database (sqlite here)')
+
+logging.info('The next 3 lines of code are the only database specific code')
+
+db = SqliteDatabase('customer.db')
+db.connect()
+db.execute_sql('PRAGMA foreign_keys = ON;') # needed for sqlite only
+
+
+logging.info('This means we can easily switch to a different database')
+
+logging.info('Enable the Peewee magic! This base class does it all')
+
+class BaseModel(Model):
+    class Meta:
+        database = db
+
+logging.info('By inheritance only we keep our model (almost) technology neutral')
+
+class Customer(BaseModel):
+    """
+        This class defines Customer, which maintains customer details.
+    """
+
+    logging.info('Specify the fields in our model, their lengths and if mandatory')
+    logging.info('Must be a unique identifier for each person')
+    id = PrimaryKeyField(primary_key = True)
+    first_name = CharField(max_length = 30)
+    last_name = CharField(max_length = 30)
+    home_address = CharField(max_length = 30)
+    phone_number = CharField(max_length = 30)
+    email_address = CharField(max_length = 30)
+    status = BooleanField()
+    credit_limit = DecimalField(max_digits = 9, decimal_places = 2)
+
+class Furniture(BaseModel):
+    """
+        This class defines Furniture, which maintains furniture details. I
+        made this just so I could see another table in the db :)
+    """
+    logging.info('The Furniture table is not going to be used')
+    id = IntegerField(primary_key = True)
+    furniture_name =  CharField(max_length = 30)
--- a/students/ericstreit/lessons/lesson05/assignment/database.py
+++ b/students/ericstreit/lessons/lesson05/assignment/database.py
@@ -1,0 +1,209 @@
+"""
+#Lesson##
+#XXXXX Exercise ##
+"""
+
+#!/usr/bin/env python3
+#import stuff here
+from pymongo import MongoClient
+import logging
+import csv
+
+
+#global variables here
+logging.basicConfig(level=logging.ERROR)
+#define function and classes
+
+class MongoDBConnection():
+    """MongoDB Connection"""
+
+    def __init__(self, host='127.0.0.1', port=27017):
+        """ be sure to use the ip address not name for local windows"""
+        self.host = host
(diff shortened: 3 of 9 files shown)


students/ericstreit/lessons/lesson04/assignment/create_customers.py
```diff
@@ -0,0 +1,54 @@
...
+        logging.info(e)
+    finally:
+        db.close()
+
+
+def customer_query(k):
+    db.connect()
+    r = Customer.select().where(Customer.first_name == k).get()
+    print(r)
+    db.close()
```
codedragon: Please use actual variable names, instead of single letters. Not only is it more difficult to tell what the variable is suppose to represent, but the debugger will be very confusing going through this code, since there are many single letter commands there.

students/ericstreit/lessons/lesson04/assignment/create_customers.py
```diff
@@ -0,0 +1,54 @@
...
+                     account__status, credit__limit):
+    """
+    Description of function
+
+    :param arg1: The first very important parameter. And a bit about
+                 what it means.
+    :param arg2: The second very important parameter. And now some
+                 description of how this is used
+    etc
+    """
```
codedragon: Very nice doc string. ;) Do get it in the habit of writing good doc strings; they are very useful.

students/ericstreit/lessons/lesson04/assignment/db_model.py
```diff
@@ -0,0 +1,56 @@
...
+    credit_limit = DecimalField(max_digits = 9, decimal_places = 2)
+
+class Furniture(BaseModel):
+    """
+        This class defines Furniture, which maintains furniture details. I
+        made this just so I could see another table in the db :)
+    """
+    logging.info('The Furniture table is not going to be used')
+    id = IntegerField(primary_key = True)
+    furniture_name =  CharField(max_length = 30)
```
codedragon: Okay, I give up. Where are the iterators and/or generators?

students/ericstreit/lessons/lesson05/assignment/database.py
```diff
@@ -0,0 +1,217 @@
...
+    creates and populates a new MongoDB database with the these data.
+    It returns 2 tuples: the first with a record count of the number of
+    products, customers and rentals added (in that order), the second with a
+    count of any errors that occurred, in the same order.
+
+    Imports CSV data into a MongoDB.
+
+    :param arg1: the name of the products csv file
+    :param arg2: the name of the customers csv file
+    :param arg3: the name of the rentals csv file
```
codedragon: This doc string is more useful. :)

students/ericstreit/lessons/lesson05/assignment/database.py
```diff
@@ -0,0 +1,217 @@
...
+                        logging.error(f'Error adding {convert_to_dict} to the rentals collection!')
+                        error_count += 1
+        except Exception as e:
+            logging.error(f'Hm, we had an error here?')
+            logging.error(e)
+
+    print(f' the number of records in the products collection is: {db.products.count()}')
+    print(f' the number of records in the customers collection is: {db.customers.count()}')
+    print(f' the number of records in the rentals collection is: {db.rentals.count()}')
+    return (record_count, error_count)
```
codedragon: This is a awful lot for one function. It could, and probably should, be broken into more functions. There is actually quite a bit of repeat code here that could be reduced.

students/ericstreit/lessons/lesson05/assignment/database.py
```diff
@@ -0,0 +1,217 @@
...
+    mongo = MongoDBConnection()
+
+    with mongo:
+        #connect to Mongo
+        db = mongo.connection.hp_norton
+
+        #name the first collection we will work on
+        products = db["products"]
+        customers = db["customers"]
+        rentals = db["rentals"]
```
codedragon: Why are you getting customers and rentals if you are not using them here?

students/ericstreit/lessons/lesson05/assignment/database.py
```diff
@@ -0,0 +1,217 @@
...
+
+    :param arg1: The product_id to search for
+    """
+    # gosh, I'm repeating myself a lot here with these Mongo commands. could
+    # be simplified into its own function, yes??
+    mongo = MongoDBConnection()
+
+    with mongo:
+        #connect to Mongo
+        db = mongo.connection.hp_norton
```
codedragon: Yes, you should not be connecting and reconnecting every time. There should be one connection when a user starts interacting with the database, and it should just stay open until they quit.

students/ericstreit/lessons/lesson05/assignment/database.py
```diff
@@ -0,0 +1,217 @@
...
+                    logging.debug(f"Adding {rented_item.get('user_id')} to list")
+            except ValueError as e:
+                logging.debug(f"skipping {rented_item.get('product_id')} due to error")
+                logging.error(e)
+
+        for name in customers.find():
+            logging.debug(f"comparing found user id to the customer list")
+            logging.debug(f"looking for {name.get('user_id')} in {customer_list}")
+            try:
+                if name.get('user_id') in customer_list:
```
codedragon: Would it be more efficient to go through the customer list to see if the user_id in the customer list is in the database, rather than going through the database looking for each customer id in the customer_list? I feel like the customer_list is likely to be quite a bit smaller, so the loop will be smaller. Same for rented_items. You can use pymongo function find_one to just look for a particular record in the database.

students/ericstreit/lessons/lesson09/assignment/database.py
```diff
@@ -0,0 +1,238 @@
...
+    def function_timer(*args, **kwargs):
+        """ nested function for timing functions """
+
+        start = time.time()
+        value = func(*args, **kwargs)
+        end = time.time()
+        runtime = end - start
+        print(f' The runtime for {func.__name__} took {runtime} seconds to complete')
+        return value
+    return function_timer
```
codedragon: Nice decorator. :) Not sure this is really what assignment 10 was going for, since we already did decorators, but more important to learn decorators then metaclasses or other meta constructs, so I'll give it to you. ;-)

students/ericstreit/lessons/lesson09/assignment/jpgdiscover.py
```diff
@@ -0,0 +1,57 @@
...
+            logging.info(f' {filename} is a directory so passing back to function using {new_dir} as the filename')
+            jpg_discovery(new_dir, png_paths)
+            png_paths.append(str(new_dir))
+        else:
+            logging.info(f'checking the file extension on {filename} for PNG')
+            if filename.endswith(".png"):
+                logging.info(f'does {filename} end with png?')
+                png_paths.append([filename])
+    # print(png_paths)
+    return png_paths
```
codedragon: Did the logging help you understand what was going on? Nice idea.

students/ericstreit/lessons/lesson09/assignment/logging_assignment/charges_calc.py
```diff
@@ -0,0 +1,171 @@
...
+        def set_level(*args, **kwargs):
+            """ nested function for timing functions """
+
+            # start = time.time()
+            value = func(*args, **kwargs)
+            # end = time.time()
+            # runtime = end - start
+            # print(f' The runtime for {func.__name__} took {runtime} seconds to complete')
+            return value
+    return set_level
```
codedragon: Decorators are run at import time, so the decorator is not going to know anything about ARGS. That is the tricky bit of this one. So, the decorator has to manipulate logging itself, and that will affect only the decorated functions, as long as you change logging before and return it back after you call the function in the decorator.

students/ericstreit/lessons/lesson09/assignment/database.py
```diff
@@ -0,0 +1,238 @@
...
+                        logging.error(f'Error adding {convert_to_dict} to the customers collection!')
+                        error_count += 1
+
+            print(f' total number of records added: {record_count}')
+            print(f' total number of documents in collection: {products.collection.estimated_data_count()}')
+        except Exception as e:
+            logging.error(f'Hm, we had an error here?')
+            logging.error(e)
+
+        return (record_count, error_count)
```
codedragon: Lol, now you really have to break up this function, if you are going to compare times of loading the different files.

students/ericstreit/lessons/lesson10/assignment/database.py
```diff
@@ -0,0 +1,241 @@
...
+https://www.blog.pythonlibrary.org/2016/05/24/python-101-an-intro-to-benchmarking-your-code/
+"""
+
+#!/usr/bin/env python3
+#import stuff here
+from pymongo import MongoClient
+import logging
+import csv
+import time
+
```
codedragon: Huh, looks like this file is in here twice.

students/ericstreit/lessons/lesson09/assignment/database.py
```diff
@@ -0,0 +1,238 @@
...
+        self.connection = None
+
+    def __enter__(self):
+        self.connection = MongoClient(self.host, self.port)
+        print(f"Establishing a connection to {self.host} on port {self.port}")
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        print(f"closing the connection to {self.host}!")
+        self.connection.close()
```
codedragon: Wait, I'm confused. Is this 9 or 10? Is adding print statements your improvement? Probably not what they had in mind.

PR #106, opened 2019-03-11
Added assignment files under lesson09 and lesson10

dtingg:
Lesson 09
Decorators: In charges_calc.py I added a decorator to the calculate_additional_fields function. I was finally able to toggle the decorator on and off from the command line by installing and using the "undecorated" module from PyPi. I'm not sure if there is a better way to do this.

Context Manager: In database.py I added some print statements to the enter and exit functions. Not sure what else I can add here.

Recursion: My png_discover.py program seems to working correctly and I formatted the results so it prints in an easy to read format.

Lesson 10
I used metaclassing in database.py to add timing information for each function.  The results are written to timings.csv.  

My findings are in conclusions.txt, but I didn't see any significant bottlenecks.  I also got all of my pytests to work.  :-)

Thank you!


codedragon (approved):
Looks good Dianna. Your functions are usually small and have one purpose, you remember to put doc strings in, and you have tests, and you seem to be understanding the difficult concepts. :) I really enjoyed having you in class. Best of luck. 
cheers,
Maria

PR #108, opened 2019-03-13
Assignment09

ktbzzz:
- Not really what how to add context manager further, it was already using it.
- There is a decorator to disable logging, but not implemented through args.


students/JaeKim/lesson09/assignment/lesson02/assignment/charges_calc.py
```diff
@@ -0,0 +1,111 @@
...
+        logging.disabled = False
+        return result
+    return logged
+
+def parse_cmd_arguments():
+    parser = argparse.ArgumentParser(description='Process some integers.')
+    parser.add_argument('-i', '--input', help='input JSON file', required=True)
+    parser.add_argument('-o', '--output', help='ouput JSON file', required=True)
+    parser.add_argument('-d', '--debug', help='debug level', required=False, default=0)
+    return parser.parse_args()
```
codedragon: The implementation through args is difficult. Fortunately, we rarely have to use args to control a decorator. :)

students/JaeKim/lesson09/jpgdiscovery.py
```diff
@@ -0,0 +1,36 @@
...
+        path = os.path.join(directory, filename)
+        if os.path.isdir(path):
+            jpegdiscovery(path, png_paths)
+        elif path.endswith(".png"):
+            current_list[1].append(filename)
+
+    if current_list[1]:
+        png_paths.extend(current_list)
+
+    return png_paths
```
codedragon: Again, remember doc strings.

codedragon (approved):
The thing I thought of for the context manager was that we always call `mongo = MongoDBConnection()` right after the context manager, so maybe that could be returned from the context manager somehow, instead of putting it all of the functions.

Code looks good. Remember to add doc strings, keep functions short, and test, test, test. :) Best of luck to you.

cheers,
Maria

PR #112, opened 2019-03-14
Lesson09 and 10 

Misseswaddles:
I figured out the decorator toggle! As a heads up, there was more detail on how to do that in the previous version of the decorator lecture. That's how I finally figured out how to do it. 
Neither part 1 or part 3 were well linted nor do they have test files written. Sorry! Once I have completed all of lesson10 I will go back and add the linting and test files. I just want to get the logic working before class ends. 
Feel free to let me know if you would prefer me to re-structure my priorities!


students/cfarris/lessons/lesson09/assignment/charges_calc_2.py
```diff
@@ -0,0 +1,141 @@
...
+            result = fn(debug, *args)
+            return result
+        else:
+            file_handler = logging.FileHandler('charges_calc.log')
+            file_handler.setFormatter(formatter)
+            logger.addHandler(file_handler)
+            logger.setLevel("INFO")
+            result = fn(debug, *args)
+            return result
+    return _decorator
```
codedragon: Shouldn't you be turning logging off after the function is run? Otherwise you are changing logging globally, not just for the functions decorated. 

Using the fact that debug is a global is definitely the easiest way for the toggle, good find.

students/cfarris/lessons/lesson09/assignment/jpgdiscover.py
```diff
@@ -0,0 +1,51 @@
...
+#use os.walk, but you can't do this with recursion.
+#may also want to look up path
+#https://docs.python.org/3/library/os.html
+
+def jpg_discovery():
+    '''
+    Though specified in the assignment, the supplied directory tree only has .png files.
+    It is therefore more pythonic to create a png_discovery.
+    '''
+    pass
```
codedragon: Lol, fair enough. :P

students/cfarris/lessons/lesson09/assignment/mdb_db.py
```diff
@@ -0,0 +1,217 @@
...
+        self.port = port
+        self.connection = None
+
+    def __enter__(self):
+        self.connection = MongoClient(self.host, self.port)
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.connection.close()
+        print("connection to MongoDB successfully closed...")
```
codedragon: You could add `db = mongo.connection.norton_products` to the context manager, since you always do that when you open the connection.

students/cfarris/lessons/lesson09/assignment/mdb_db.py
```diff
@@ -0,0 +1,217 @@
...
+                   collection.count(),
+                   time_to_run)
+        print(collection.name,
+              " finished: ",
+              num_processed,
+              count_before_running,
+              collection.count(),
+              time_to_run)
+        data_list.append(results)
+        return data_list
```
codedragon: There is way too much in this context manager, even the return statement is in the context manager. Try to do just the minimum, and then close the connection. There is also too much in this function. Notice that your doc string says this function is doing two things, returning a dictionary and saving data, so this should probably be two functions. Sorry if I am repeating myself, never sure what I've said in past reviews. 🤷‍♀️

students/cfarris/lessons/lesson09/assignment/test_mdb_db.py
```diff
@@ -0,0 +1,152 @@
...
+        self.host = host
+        self.port = port
+        self.connection = None
+
+    def __enter__(self):
+        self.connection = MongoClient(self.host, self.port)
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.connection.close()
```
codedragon: You don't have to re-create this context manager in every file, you can just import it.

students/cfarris/lessons/lesson10/assignment/databse.py
```diff
@@ -0,0 +1,252 @@
...
+        self.port = port
+        self.connection = None
+
+
+    def __enter__(self):
+        self.connection = MongoClient(self.host, self.port)
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.connection.close()
```
codedragon: I don't think it is very useful to time the connection to the mongo database, but yes, this is how you would do it.

codedragon (approved):
Nicely done, Carol. Made a few comments, but some are probably repeat. Very much enjoyed having you in class again, and feel free to do pull requests if you want me to look at changes. 
cheers,
Maria

PR #114, opened 2019-03-15
submit lesson08 hw

ShujinCao:


students/ShujinCao/lesson08/assignment/inventory.py
```diff
@@ -0,0 +1,25 @@
...
+        row = customer_name, item_code, item_description, item_monthly_price
+        writer.writerow(row)
+
+def single_customer(customer_name, invoice_file):
+    def within_function(target_csv):
+        with open(invoice_file) as source_csv:
+            original_entries = [line.split(",") for line in source_csv if line]
+            add_furniture(customer_name, item[1], item[2], item[3])
+
+    return within_function
```
codedragon: I don't think this can work. add_furniture takes 5 input variables, but you are only giving it 4 in your closure. The idea was to use functools.partial to change the signature from 5 to 3. And then you have a variable `orignal_entries` that you are not using, and where did `item` come from?

codedragon (approved):
You have a closure, unfortunately, it doesn't work. Tests would probably be helpful to make sure it is doing what it is suppose to be doing. Remember to add doc strings, so the user knows how to use the closure.

PR #115, opened 2019-03-16
submit lesson09 HW

ShujinCao:


students/ShujinCao/lesson09/assignment/charges_calc.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def decorated_func(*args, **kwargs):
+        logging.disable(logging.CRITICAL)
+        print("some logging has been disabled")
+        returned_value = func(*args, **kwargs)
+        logging.disable(logging.NOTSET)
+        print("some logging has been re-enabled")
+        return returned_value
+
+    return decorated_func
```
codedragon: Decorator looks good. Doesn't toggle from the argparse arguments, but that was not clear in the assignment, tricky to do, and seldom necessary.

students/ShujinCao/lesson09/assignment/database.py
```diff
@@ -0,0 +1,78 @@
...
+        self.port = port
+        self.connection = None
+
+    def __enter__(self):
+        self.connection = MongoClient(self.host, self.port)
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.connection.close()
+
```
codedragon: Not sure exactly what is different in this file. I was thinking something useful to add to the MongoDBConnection context manager would be getting `mongo.connection.media` in there, since it is always run after the context manager is opened.

students/ShujinCao/lesson09/assignment/jpgdiscover.py
```diff
@@ -0,0 +1,20 @@
+import os
+
+def jpegdiscovery(directory, png_paths=None):
```
codedragon: Don't forget your doc strings!

codedragon (approved):
You are good Shujin, although if you want to turn in 10, I will be glad to review it. Please keep in mind to create small functions, write doc strings, and test everything. :) I enjoyed having you in class. Best of luck in the future.
cheers,
Maria

PR #116, opened 2019-03-16
my assignment10_update

jbyamindi:
Here is my assignment 10, It can run but It can collect data. Maybe something wrong with my Pewee.
Feel free to check it for me.
Thanks


codedragon (approved):
Hello Jean-Babtiste,
So, the metaclass looks fine, but it looks like you were trying to use it as a decorator instead of a metaclass.

PR #117, opened 2019-03-16
Assignment 10

CodeChameleon:
I put all the functions into a new class so that it could use the new meta class. I was able to get the meta class to work. I just loops through all of the attributes and assigns a decorator function to any methods (that don't start with _).

I used the updated version of database.py from the concurrent version because I had already broken the functions into smaller bits in that version.

I didn't notice any serious timing issues. I ran data imports that have 30; 1500; 2000; and 10,000 records. The latter took 0.3s. Perhaps if I did a million records it would start slowing down.

### DIFF
--- a/students/DrewSmith/lessons/lesson10/assignment/databse.py
+++ b/students/DrewSmith/lessons/lesson10/assignment/databse.py
@@ -1,0 +1,199 @@
+'''
+HP Norton inventory and customer information database functions
+'''
+
+
+import csv
+import pathlib
+import pymongo
+import logging
+import types
+import inspect
+from datetime import datetime
+from pathlib import Path
+from pymongo import MongoClient
+import config as config
+from multiprocessing.pool import ThreadPool
+
+TIMING_PATH = "timing.csv"
+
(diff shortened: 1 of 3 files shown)


students/DrewSmith/lessons/lesson10/assignment/databse.py
```diff
@@ -0,0 +1,204 @@
...
+    :param counts: reference to counts dictionary to log records affected
+    '''
+    def fncomposite(*args, **kwargs):
+        ''' Timing function called on decorated functions '''
+        start = datetime.now()
+        result = func(*args, **kwargs)
+        end = datetime.now()
+        # Write to log file
+        with open(TIMING_PATH, "a") as file:
+            file.write(f"{func.__name__},{(end - start)},{counts[func.__name__]}\n")
```
codedragon: Nicely done. Some had a difficult time getting the counts in the decorator, this is a good way to do it.

codedragon (approved):
Looks good Drew. It was a pleasure having you in class. Best of luck in the future.
cheers,
Maria

